PR #20, opened 2015-01-25
Allow arbitrary byte-array for secret

zerodivide1:
The current implementation does not allow using an arbitrary secret (such as a random generated secret that is base64-encoded).

The changes I have added allow using an arbitrary array of bytes as the secret.


pose:
@zerodivide1 Thanks for your contribution! In a couple of minutes it will be published in the 2.1.0 version.

zerodivide1:
@pose Thank you for taking the merge!

Issue #18, opened 2015-01-02
Repackaging Jackson package breaks Jackson annotations


kucharzyk:
Repackaging Jackson package breaks Jackson annotations :(

When I serialize user claim using JwtSigner the following code is not working.

```java
public class UserAuthority  {
    @NotNull
    @ManyToOne
    @com.fasterxml.jackson.annotation.JsonIgnore
    private User user;
}
```
Annotation JsonIgnore is ignored  and this field is serialized (causing infinite loop).
When I switch to  annotations from com.auth0.jwt.internal.com.fasterxml.jackson everything works perfectly.

```java
public class UserAuthority  {
    @NotNull
    @ManyToOne
    @com.auth0.jwt.internal.com.fasterxml.jackson.annotation.JsonIgnore
    private User user;
}
```

Is repackaging com.fasterxml.jackson and org.apache.commons.codec absolutely necessary.
Can't You add them as standard dependencies?

pose:
Hi @hexangular,

Unfortunately, people using our library could be using a different Jackson version (this was the case why we decided to merge this change, see https://github.com/auth0/java-jwt/pull/6). I agree is not elegant repackaging but how would you address having different versions of Jackson? Do you think there is a better and more elegant way to tackle the issue?

Thanks in advance,

hexangular:
Hi @pose ,

Maybe can You provide some kind of adapter for users and remove Jackson from dependencies.
Adapter interface would have all serialization/deserialization methods and user would implement it using his version of Jackson or maybe even using different library.

It would require a liite more work from users but I think It can be a solution.

What do You think?

pose:
Hi @hexangular,

One of the guiding lights of java-jwt is making the library simple to use. That means that we should provide a default JSON serialization/deserialization mechanism in the library. I agree with your point of creating an Adapter Interface as long as is clear and doesn't add much complexity or too many abstractions but by default it should work as it does now.

Maybe there is another way to solve this: What if we use optional dependencies in Maven? I'm thinking something like, if you include GSON or your own JSON library we use that but if we don't find it we switch to our default. One of the drawbacks would be that those checks could be expensive as they use reflection. Thoughts?

thaggie:
Or you could provide multiple binaries and let consumers decide whether or not to have things repackaged.

cowtowncoder:
Couple of comments here, in case they might help.

[...]

PR #24, opened 2015-04-01
Replace jackson-databind with jodd-json

klette:
Replaces jackson-databind with [jodd-json].

The motivation behind this is to reduce the likeliness of
a version conflict in existing applications. Jackson is
one of the most common JSON libraries out there, but it has
some severe issues when multiple versions are available on the
classpath.

jodd-json is a small JSON parser/serializer library with a minimal
footprint. The resulting shaded jar of java-jwt is about half the size
compared to when its bundled with jackson-databind.

The patch has a bad side effect in that it created different tokens
than the jackson-bind implementation, since it escapes all the
characters defined in RFC4627, while jackson-databind only escapes
the MUST characters. This is visible in the test cases that include
"/" in the claim values.

Another "bad" effect is the use of Map<String, Object>, and Object
references. I've tried to add the needed type checks, but there
might dragons left to slay.

This commit should fix issue #18 and similar multiple jackson versions
issues that might be out there is the wild.

The pull request needs a fair bit a review, as I'm not that familiar with either JWT or this particular implementation of it. 

[jodd-json]: http://jodd.org/doc/json/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,9 +47,9 @@
   <dependencies>
     <!-- For JWT parsing-->
     <dependency>
-      <groupId>com.fasterxml.jackson.core</groupId>
-      <artifactId>jackson-databind</artifactId>
-      <version>2.0.1</version>
+      <groupId>org.jodd</groupId>
+      <artifactId>jodd-json</artifactId>
+      <version>3.6.5</version>
     </dependency>
     <dependency>
       <groupId>commons-codec</groupId>
--- a/src/main/java/com/auth0/jwt/JWTAudienceException.java
+++ b/src/main/java/com/auth0/jwt/JWTAudienceException.java
@@ -1,30 +1,29 @@
 package com.auth0.jwt;
 
-import com.fasterxml.jackson.databind.JsonNode;
 
 import java.util.ArrayList;
 import java.util.List;
 
 public class JWTAudienceException extends JWTVerifyException {
-    private JsonNode audienceNode;
+    private Object audienceNode;
 
-    public JWTAudienceException(JsonNode audienceNode) {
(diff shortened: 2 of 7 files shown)


klette:
This is one approach of using something smaller than Jackson to do the JSON-lifting. The needs of java-jwt json wise is quite limited, so a smaller library might suit the need quite well. I'm sure there are other JSON serializer that we might look into if this one seems not to fit the bill quite right :)

It shouldn't be to hard to roll our own actually, if that is preferred.

lordofthejars:
Even you could use JSON Processing API (but only the jar of the spec) so everyone can choose the implementation.  It would be a light dependency from the point of view of project, and in cases you are using any application server it will comes with an implementation by default.

pose:
Thanks @klette @lordofthejars. The ideal would be to adopt JSR-353 but AFAIK none of the main JSON libraries support that spec (Jackson, GSON). ([source](http://stackoverflow.com/questions/27982463/do-jackson-and-gson-directly-implement-the-standard-jsr-353)). 

Concerning `jodd-json` we have decided that for now we will keep using Jackson but we might revisit this in the future.

cowtowncoder:
Another possibility might be jackson-jr:

     https://github.com/FasterXML/jackson-jr

which is relatively small, and when using repackaged version (jr-all) does not conflict in any way with full Jackson. It does use high-performance streaming parser/generator part of Jackson, but otherwise implements a bare-bone databinding to/from Beans and "untyped" containers (Maps, Lists, wrappers).
The intent here is to create something that is bit more streamlined for embedded use, where use of annotations is not needed not custom extensibility; but where good performance and support for things like generics (typed Collections, Maps) is useful.

Issue #19, opened 2015-01-23
Default charset used all over the place


maksimlikharev:
Can you guys put charset argument in all places you use: getBytes(), so it will be predictable at least what I sign.

thaggie:
I've created a pull request for this: https://github.com/auth0/java-jwt/pull/25

lbalmaceda:
Hello @maksimlikharev. 
Please try v3 as this issue it's already fixed there.
Cheers.

PR #25, opened 2015-04-02
Fix for #19 updated all calls to getBytes or new String to pass a C...

thaggie:
...harset of UTF-8


pose:
Thanks @thaggie. This will be merged and added in the next major version as it's a potentially breaking change. Afterwards, we will add overloads in each method to allow specifying the charset manually and using `utf-8` as the default.

Orbiter:
I suggest https://github.com/auth0/java-jwt/pull/49 instead as StandardCharsets.UTF_8 is a standard java constant now

PR #26, opened 2015-04-02
JWTVerifier was creating a URL safe base 64 decoder but wasn't using it ...

thaggie:
...as it was using the non URL safe static methods on Base64.

I'm assuming here that these things are wanted to be decoded URL safe.


pose:
Good catch. Thanks @thaggie

PR #27, opened 2015-04-02
Moved the ObjectMapper and algorithms to be static as they're not needed...

thaggie:
... / used on a per instance basis and are safe to be static and so avoid the allocation overhead at contruction time.


pose:
Thanks @thaggie. For the moment, we will keep ObjectMapper non-static to potentially allow reconfiguring it in the future. We agree that the algorithms should be static, but in this case they should be stored in an immutable map to ensure thread safety. Could you make these changes so we can get this merged?

PR #29, opened 2015-04-22
fix(test): Fix race condition in  RoundtripTest.shouldOptionsIat

### LINKED ISSUE #28: RoundtripTest.shouldOptionsIat test has a race condition
This is extremely minor and rare, but if the encode and verify happen across a second boundary, the test can fail asserting `iat >= before / 1000l`

christopherthielen:
closes #28


pose:
Thanks @christopherthielen!

PR #30, opened 2015-04-22
Remove dependency on commons-codec.

christopherthielen:
I see there's already some discussion regarding whether or not it's a good idea to keep Jackson as a dependency.  Considering it currently is a dependency, however, it's worth noting that Jackson also provides Base64 en/decoding.  This PR removes the commons-codec dependency and replaces the en/decode calls with a small Base64 class that delegates to Jackson.

All current tests pass.  

I'm packaging and deploying original-java-jwt.jar (no dependencies rolled in via shade) into my app.  I don't really want a dependency on commons-codec, thus the motivation for this.  It's just a thought :+1:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,12 +52,6 @@
       <version>2.0.1</version>
     </dependency>
     <dependency>
-      <groupId>commons-codec</groupId>
-      <artifactId>commons-codec</artifactId>
-      <version>1.4</version>
-    </dependency>
-
-    <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.11</version>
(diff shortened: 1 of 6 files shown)


pose:
Thanks @christopherthielen. We like the idea of replacing codecs dependency but we need to make some checks in order to see if it works as expected.

schmitch:
Is this really necessary?! I mean Java8 has a Base64 decoder, so basically people could just use it and force a Version which only works on Java8

sibay:
i saw this pr too late. i just added a new pr replacing commons-codec and jackson with jdk8 base64 and boon json. Someone should decide what to take.

PR #32, opened 2015-05-27
Update README.md

nicholasle:
Update the sample application to make it works for Auth0's users.
Update maven to the latest version of java-jwt 2.1.0


pose:
@nicholasle thanks!

PR #43, opened 2015-10-19
JWT Vulnerability Fix - Force verification algorithm

KSanthanam:
This fixes the JWT vulnerability highlighted by Tim McClean in this article https://auth0.com/blog/2015/03/31/critical-vulnerabilities-in-json-web-token-libraries/

### DIFF
--- a/src/main/java/com/auth0/jwt/JWTVerifier.java
+++ b/src/main/java/com/auth0/jwt/JWTVerifier.java
@@ -2,6 +2,7 @@ package com.auth0.jwt;
 
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.lang.IllegalArgumentException;
 import java.security.InvalidKeyException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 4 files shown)


src/test/java/com/auth0/jwt/JWTVerifierTest.java
```diff
@@ -25,27 +25,27 @@ public void constructorShouldFailOnEmptySecret() {
```
Rebzie: I think you should write a unit test for forcing algorithm

src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -78,9 +79,12 @@ public JWTVerifier(byte[] secret) {
      * @throws JWTVerifyException    when expiration, issuer or audience are invalid
```
Rebzie: update java doc

Rebzie:
code looks good see comments on unit test and java doc

rolodato:
Hi @KSanthanam, thanks for your pull request. However, this library isn't affected by this vulnerability because it only supports symmetric signatures: https://github.com/auth0/java-jwt/blob/master/src/main/java/com/auth0/jwt/JWTVerifier.java#L55-L58

Eventually we'd like to add support for asymmetric signatures, so we could merge this into a branch other than `master`. Could you edit this PR so that it targets the `rs` branch instead? We're not ready to start development on this, but your PR would be a good starting point. Thanks!

pose:
@KSanthanam Also, please take into account that if you find a security issue you should follow the procedure detailed here: auth0.com/whitehat Thanks!

KSanthanam:
@rolodato Thank you. Yes, I noticed it doesn't do RSA. Is anyone working on that? Shall I give it a crack?

rolodato:
@KSanthanam: Not right now, but feel free to change this PR so that it targets the `rs` branch instead of `master`. That way we can use this as a base when we start development.

silhouettes:
Two things:

1. This library now supports asymmetric signatures, and it is in master.
2. However, the vulnerability is prevented in a different way (see #53 for details).

So, it would still be nice to be able to whitelist acceptable algorithms, but this PR is significantly out of date, and not being able to do so does not seem to be critical.

PR #49, opened 2016-02-11
StandardCharsets.UTF_8 instead of "UTF-8" string lookup

Orbiter:
this should make it a bit faster


hzalaz:
Thanks for the PR, we enforce UTF-8 in v3 so we'll close this one

PR #51, opened 2016-03-01
Update to documentation with how to generate a JWT

christophe-f:


christophe-f:
You might also want to update the following page https://auth0.com/docs/api/v2/tokens to add a link to this repo.

Currently you have:
"An Auth0 APIv2 token can be generated on the Auth0 APIv2 explorer page or it can be created programmatically by building the JWT, including the desired scopes, and signing it with the tenant api key/secret."

But no example how to create programmatically.

PR #72, opened 2016-09-05
Add param Header in sing function

gasmuro1:
Hi 

I need add the headers param, to adds diferents headers , like the pyjwt python library 

Regards


hzalaz:
Thanks for the PR. We added this feature in v3 (we'll release it soon)

PR #79, opened 2016-10-13
Upgrade commons-io

postalservice14:
Your commons-io version is very old and is conflicting with other libraries we have.  Any chance for an upgrade?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
     <dependency>
       <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
-      <version>1.3.2</version>
+      <version>2.5</version>
     </dependency>
 
     <!-- test -->


hzalaz:
@postalservice14 we'd need to check if something breaks first. I believed the dependencies where bundled in the same jar for this library so how it's conflicting with other libraries?

postalservice14:
In maven `auth0-spring-security-api:0.3.0` requires `java-jwt:2.2.0` which requires `org.apache.commons:commons-io` which is conflicting with another library that requires `commons-io:commons-io:2.5`

For our tests to pass and our application to work we had to add an exclusion in the pom like this:
```xml
<dependency>
    <groupId>com.auth0</groupId>
    <artifactId>auth0-spring-security-api</artifactId>
    <version>0.3.0</version>
    <exclusions>
        <exclusion>
            <groupId>org.apache.commons</groupId>
            <artifactId>commons-io</artifactId>
        </exclusion>
    </exclusions>
</dependency>
```

I'm not entirely sure what you're asking, but what I do know is that the old dependency on this library is causing issues for us.  Also, I guess it goes without saying, that `auth0-spring-security-api` will need to be updated to pull in the new version of `java-jwt` as well for it to fully solve our problem.

hzalaz:
We are using the shade plugin to repackage the dependencies inside a java-jwt jar https://github.com/auth0/java-jwt/blob/master/pom.xml#L115 (not something I'd like but it's how it's built) so it should not fetch them and break your build.

However, recently there was a PR to use change the commons-io groupId and that was not updated in the plugin. Will try to test the new commons-io version next week.

postalservice14:
Ahh, yeah, I see, the groupId was changed the shade plugin not work.  Looks like your shade plugin is still looking for `org.apache.commons.io`.

Thanks.

adrogon:
@postalservice14 This has already been changed in java-jwt:2.2.1, but they haven't released an auth0-spring-security-api:0.3.1 that would contain that dependency change.
What you can do is add a dependency to java-jwt:2.2.1 in your project, or wait 0.3.1 to be released.

hzalaz:
@adrogon the PR updates commons-io to version 2.5

hzalaz:
will close this one and update the shade plugin to avoid the clash.

PR #80, opened 2016-10-13
Fix problem with thread safety that produces a race condition

postalservice14:
We ran into a scenario where there was a race condition to Base64 decode tokens at the same time.  Since Base64 was an instance on the verifier class, the call to `Base64.reset` on the `decode` function was clearing the same instance.

I hope that makes sense.  Race conditions are hard to write a test for....I tried, but couldn't get it to fail. :-/

### DIFF
--- a/src/main/java/com/auth0/jwt/JWTVerifier.java
+++ b/src/main/java/com/auth0/jwt/JWTVerifier.java
@@ -33,7 +33,6 @@ public class JWTVerifier {
     private PublicKey publicKey;
     private final String audience;
     private final String issuer;
-    private final Base64 decoder = new Base64(true);
 
     private final ObjectMapper mapper;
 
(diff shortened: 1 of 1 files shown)


postalservice14:
This is a real problem for us.  Currently, we had to build our own `java-jwt` while we wait for this to get pull in....  Any chance we can get this pulled in?

Let me know if you have any questions.

arcseldon:
@postalservice14 - which version of this library are you using?  We recently applied an [update](https://github.com/auth0/java-jwt/commit/559da8412bc7ea84ca5d4e1692e9c0669573fbf1) to bump `commons-codec` from `1.4` to `1.10` - there was a known bug with v1.4.  Checking the 1.10 JavaDoc it does state that [Base64 is ThreadSafe](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html) - can you confirm please you are using version `2.2.1` of this library and you only have commons-codec v1.10 in your classpath?

postalservice14:
I think that may have fixed it.  I'll reopen if I run across it again.  Thanks!

PR #81, opened 2016-10-13
Update Shade to use commons-io instead of org.apache.commons.io

postalservice14:
commons-io appears to have had it's groupId changed from `org.apache.commons.io` to `commons-io`, but shade was not updated.


postalservice14:
Sorry, I guess this isn't in issue after all.  Looks like `commons-io` groupId uses the same package.

Issue #117, opened 2016-12-09
Array claim [3.0.1]


chadjaros:
I'm trying to update from 2.1.0 to 3.0.1, but it seems like it's illegal to add a claim which consists of an array of values. Is this intentionally restricted, and is that restriction present in the [RFC](https://tools.ietf.org/html/rfc7519)?

chadjaros:
From the RFC:

    2. Terminology
    ...    
    Claim Value [Page 5]
        The value portion of a claim representation.  A Claim Value can be
        any JSON value.

It seems that Arrays and Maps should be allowed parameters in `JwtCreator#withClaim`

hzalaz:
@chadjaros it can be added. If you want a faster turnaround please submit a PR and we'll review and merge

lbalmaceda:
Available in version `3.1.0`.

PR #123, opened 2016-12-14
Accept Array type on verification and creation.

lbalmaceda:
Now we can specify Arrays of type String or Integer for creation/verification of JWTs.

1. Get a Claim as a custom class:

```java
Claim claim = jwt.getClaim("user");
UserProfile user = claim.as(UserProfile.class);
```

2. Create a token with an Array Claim:

```java
String jwt = JWTCreator.init()
                .withArrayClaim("name", new String[]{"a", "b", "c"})
```

3. Verify that a token contains an Array Claim with at least specific items:

```java
DecodedJWT jwt = JWTVerifier.init(Algorithm.HMAC256("secret"))
                .withArrayClaim("name", "a", "b")
                .build()
                .verify(token);
```

Note that in this example, `UserProfile` class must have a no-args public constructor for deserialization to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,8 +286,6 @@ JWT.require(Algorithm.HMAC256("secret"))
     .verify("my.jwt.token");
 ```
 
-> The value of the custom Claim in all the cases must be of a `Integer`, `Double`, `Date`, `String`, or `Boolean` class.
-
 
 ### Claim Class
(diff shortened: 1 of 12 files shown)


lbalmaceda:
Adds the features discussed in https://github.com/auth0/java-jwt/issues/117

Issue #118, opened 2016-12-09
Clock is package private [3.0.1]


chadjaros:
Should the Clock class, and JWTVerifier#build method be public instead of package private? Making these package private doesn't do me any good if I want to use a custom clock for unit testing.

chadjaros:
Related to #50

hzalaz:
Might be a good idea. Will have to review the API to make this possible. PRs are always welcomed, we were busy handling other work but we'll try to be up to date with them

lbalmaceda:
Available in version `3.1.0`.

Issue #50, opened 2016-02-29
Allow creation and verification for a specified time


chadjaros:
Unit testing classes with expiring tokens with this library is difficult because it is tied to `System.currentTimeMillis()` in multiple places. I use JodaTime, which has a number of features which make unit testing easier, including freezing the joda clock at a specified time. I would like to inject the time for iat/exp on creation, and for verifying exp, such that it can be made to work with this Joda functionality. 

I am already setting iat/exp manually to leverage this feature, but it doesn't seem possible to do so for verification at this time.

arcseldon:
@chadjaros - do you have a PR?

chadjaros:
I do not. I set up a workaround in my unit tests because that was the path of lesser resistance.

hzalaz:
@chadjaros the path of lesser resistance is to make clock public and figure out how to inject it and avoid people messing with this since it's meant only for testing.  So if you are able please submit a PR for allowing to define a clock

PR #125, opened 2016-12-14
Make Clock customization accessible for verification

lbalmaceda:
Now when working with time related claims, the JWTVerifier allows to specify the custom `Clock` instance used on the verification. We want this to be as hidden as possible for the normal user. Devs can access it by performing a cast.

```java
BaseVerification verification = (BaseVerification) JWT.require(Algorithm.RSA256(key))
    .acceptLeeway(1)
    .acceptExpiresAt(5);
Clock clock = new CustomClock(); //Must implement Clock interface
JWTVerifier verifier = verification.build(clock);
```

Related to: https://github.com/auth0/java-jwt/issues/50 https://github.com/auth0/java-jwt/issues/118

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ If you need to test this behaviour in your lib/app cast the `Verification` insta
 BaseVerification verification = (BaseVerification) JWT.require(Algorithm.RSA256(key))
     .acceptLeeway(1)
     .acceptExpiresAt(5);
-Clock clock = new Clock();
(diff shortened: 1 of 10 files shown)


lbalmaceda:
The last commit can be avoided, as it only extracts a single method interface for the `Clock` class. This can be set from a new constructor receiving the `Date` value to be returned in the `clock.getToday()` call.

Issue #132, opened 2017-01-16
Timestamps are limited by Integer/int to 2038-01-19T04:14:07.000+0100


larspettern:
If the "expires at" claim is set to a date larger than 2038-01-19, then it gets broken by the fact that the date is translated into a int. See [PayloadSerializer line 58](https://github.com/auth0/java-jwt/blob/master/lib/src/main/java/com/auth0/jwt/impl/PayloadSerializer.java)

The latest date possible to represent in seconds as a Java Integer is 

`new Date(Integer.MAX_VALUE * 1000L);`

Which corresponds to 2038-01-19T04:14:07.000+0100

The date claims should always be respresented as Long in Java since the second values will overflow the Integer max value and you will get an erroneous value.

This applies to the other timestamp claims also (issued at, and not before) together with the custom date claims.

lbalmaceda:
@larspettern thanks for the report. I'll look into it 👍

PR #137, opened 2017-01-23
Fix Date value conversion

lbalmaceda:
Bug fix for the Date value parsing.
Related https://github.com/auth0/java-jwt/issues/132

### DIFF
--- a/lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
+++ b/lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
@@ -220,6 +220,19 @@ public class PayloadDeserializerTest {
     }
 
     @Test
+    public void shouldGetLargeDateWhenParsingNumericNode() throws Exception {
+        Map<String, JsonNode> tree = new HashMap<>();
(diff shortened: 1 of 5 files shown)


lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
```diff
@@ -220,6 +220,19 @@ public void shouldGetDateWhenParsingNumericNode() throws Exception {
...
+    public void shouldGetLargeDateWhenParsingNumericNode() throws Exception {
+        Map<String, JsonNode> tree = new HashMap<>();
+        long seconds = Integer.MAX_VALUE + 10000L;
+        LongNode node = new LongNode(seconds);
+        tree.put("key", node);
+
+        Date date = deserializer.getDateFromSeconds(tree, "key");
+        assertThat(date, is(notNullValue()));
+        assertThat(date.getTime(), is(seconds * 1000));
+        assertThat(date.getTime(), is(2147493647L * 1000));
```
nikolaseu: `Integer` is always 32 bits on java?
it's weird how java makes you explicitly cast to `(Object)` sometimes, but when you sum a long to an int it doesn't care.

PR #138, opened 2017-01-23
Add Key Id setter and set JWT Type after signing

lbalmaceda:
Add setter for header's 'kid' value. Also set the 'typ' to "JWT" after the signing process.

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -74,7 +74,7 @@ public final class JWTCreator {
         }
 
         /**
-         * Add a specific Issuer ("iss") claim.
+         * Add a specific Issuer ("iss") claim to the Payload.
          *
          * @param issuer the Issuer value.
          * @return this same Builder instance.
@@ -85,7 +85,7 @@ public final class JWTCreator {
         }
 
         /**
-         * Add a specific Subject ("sub") claim.
+         * Add a specific Subject ("sub") claim to the Payload.
          *
          * @param subject the Subject value.
          * @return this same Builder instance.
@@ -96,7 +96,7 @@ public final class JWTCreator {
         }
 
         /**
-         * Add a specific Audience ("aud") claim.
+         * Add a specific Audience ("aud") claim to the Payload.
          *
          * @param audience the Audience value.
          * @return this same Builder instance.
@@ -107,7 +107,7 @@ public final class JWTCreator {
         }
 
         /**
-         * Add a specific Expires At ("exp") claim.
+         * Add a specific Expires At ("exp") claim to the Payload.
          *
          * @param expiresAt the Expires At value.
          * @return this same Builder instance.
@@ -118,7 +118,7 @@ public final class JWTCreator {
         }
 
         /**
-         * Add a specific Not Before ("nbf") claim.
+         * Add a specific Not Before ("nbf") claim to the Payload.
          *
          * @param notBefore the Not Before value.
          * @return this same Builder instance.
@@ -129,7 +129,7 @@ public final class JWTCreator {
         }
 
         /**
-         * Add a specific Issued At ("iat") claim.
+         * Add a specific Issued At ("iat") claim to the Payload.
          *
          * @param issuedAt the Issued At value.
          * @return this same Builder instance.
@@ -140,7 +140,7 @@ public final class JWTCreator {
         }
 
         /**
-         * Add a specific JWT Id ("jti") claim.
+         * Add a specific JWT Id ("jti") claim to the Payload.
          *
          * @param jwtId the Token Id value.
          * @return this same Builder instance.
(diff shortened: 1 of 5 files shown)


lib/src/main/java/com/auth0/jwt/JWTCreator.java
```diff
@@ -33,6 +35,7 @@ private JWTCreator(Algorithm algorithm, Map<String, Object> headerClaims, Map<St
             SimpleModule module = new SimpleModule();
             module.addSerializer(ClaimsHolder.class, new PayloadSerializer());
             mapper.registerModule(module);
+            mapper.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true);
```
nikolaseu: I guess this is required because of the JsonMatcher and it's own implementation of JSON serialization?
Maybe it was just better to use jackson there? It would have been, at least, less code.

lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
```diff
@@ -28,15 +30,29 @@ public void shouldThrowWhenRequestingSignWithoutAlgorithm() throws Exception {
...
+    @Test
+    public void shouldAddKeyId() throws Exception {
+        String signed = JWTCreator.init()
+                .withKeyId("56a8bd44da435300010000015f5ed")
+                .sign(Algorithm.HMAC256("secret"));
+
+        assertThat(signed, is(notNullValue()));
+        String[] parts = signed.split("\\.");
+        String headerJson = new String(Base64.decodeBase64(parts[0]), StandardCharsets.UTF_8);
+        assertThat(headerJson, JsonMatcher.hasEntry("kid", "56a8bd44da435300010000015f5ed"));
```
nikolaseu: since this code is repeated a few times, maybe we could add a matcher, something like `JWTMatcher.hasHeaderEntry` ? Or at least a method to wrap the split, base64decode of parts[0] and the hasEntry.

lib/src/main/java/com/auth0/jwt/JWTCreator.java
```diff
@@ -261,6 +275,7 @@ public String sign(Algorithm algorithm) throws IllegalArgumentException, JWTCrea
                 throw new IllegalArgumentException("The Algorithm cannot be null.");
             }
             headerClaims.put(PublicClaims.ALGORITHM, algorithm.getName());
+            headerClaims.put(PublicClaims.TYPE, "JWT");
```
nikolaseu: just saw this, but why the `PublicClaims.TYPE`, `PublicClaims.ALGORITHM`? isn't this making the code harder to understand? it's a jwt lib, the field is `typ` so I would have expected something like `PublicClaims.TYP`
At least me, as an end-user, would not be really confident and would have to check the code to see if it's the actual value I'm looking for.

lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
```diff
@@ -158,43 +187,43 @@ public void shouldAcceptCustomClaimOfTypeString() throws Exception {
         String jwt = JWTCreator.init()
                 .withClaim("name", "value")
                 .sign(Algorithm.HMAC256("secret"));
-        String token = "eyJhbGciOiJIUzI1NiJ9.eyJuYW1lIjoidmFsdWUifQ.4qDWJcNQHDVDW1iAcIgZNiu-qqJQ0RIq8X3ETijBx5k";
 
         assertThat(jwt, is(notNullValue()));
-        assertThat(jwt, is(token));
+        String[] parts = jwt.split("\\.");
+        assertThat(parts[1], is("eyJuYW1lIjoidmFsdWUifQ"));
```
nikolaseu: why are these tests different from the others (the header ones)? I mean, here you have the hardcoded value, and in the others you decode them.

lib/src/main/java/com/auth0/jwt/JWTCreator.java
lbalmaceda: This was the previous solution to the JSON tests that were matching against a full json string. The issue was with java 7 and 8 were the json entries were ordered in a different way. After I added the entry/value check instead of comparing against a full json output, I think we can remove this.

nikolaseu: I meant the JsonMatcher could use jackson instead of having it's own implementation of {object/list/array/map}ToString

lbalmaceda: Yes, I agree. There was a reason though of why I avoided that approach at first. I guess I wanted to assure that the output keys/names were actually the ones the API was expecting, mostly because there is inconsistence between the naming of the variables and the json keys.

lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
lbalmaceda: Because of the JSON entries ordering issue: In this case because the decoded body contains just 1 key, I don't have issues with the JSON order and the base64 encoded part is the same both for Java 7 and Java 8. In the case of the header, besides the key I'm specifying in each test, the signing process adds the `typ` and `alg` claims. So entries order can mess things a bit and return a different base64 encoded string for the same (but in different order) values.

lib/src/main/java/com/auth0/jwt/JWTCreator.java
lbalmaceda: I know the class is public (and also it's name, lol) but they are not really intended to be used by the user. There are individual setters for each of this "public claims" in the `JwtCreator` and `JwtVerifier` classes for that purpose.

PR #141, opened 2017-02-22
Enable parsing JSON with newline postfix

xelat09:
Couldn't we say JSON with leading/trailing whitespace is valid too? It solves my issues with trailing newlines..


hzalaz:
@xelat09 thanks for the PR but I am not sure that this is something I'd like. Reviewing this part of the code we don't need to pre-validate the String we pass to decode and let the parser decide if its valid or not. Probably the parser will allow white spaces so I'll close this and leave that kind of validation to the JSON parser

PR #143, opened 2017-02-26
Enable parsing JSON with newline postfix

xelat09:
another try :)
yes, the jackson parser allows white spaces. the null check allows a consistent exception handling, so it might have a value..


xelat09:
oops, i broke something :), well you know the request., i'll pull this back... someone also reported the issue

ty 4 the lib

Issue #146, opened 2017-03-08
Allow returning claims or specific claim as a Map


hzalaz:
We need to add to `Payload` interface a method like 

```java
Map<String, Object> jwt.getClaimsAsMap();
```

That will allow to get all claims as a simple map, useful for cases where the claim payload is way too complex to parse with `Claim` class.

```java
JWTVerifier verifier = JWT.require(Algorithm.HMAC256("secret"))
        .withIssuer("auth0")
        .build(); //Reusable verifier instance
JWT jwt = verifier.verify(token);
Map<String, Object> claims = jwt.getClaimsAsMap();
```

Also additionally we can add the same feature to the `Claim` object like

```java
JWT jwt = verifier.verify(token);
Claim claim = jwt.getClaim("user_metadata");
Map<String, Object> metadata = claim.getAsMap();
```

### Considerations

- Will always use the default JSON mapper config so if there is a structure that needs a custom mapping it will fail or have undefined behaviour
- There will be no hook to register custom mappings
- Only available to decode tokens not for signing

PR #152, opened 2017-03-14
Allow to get a Claim as Map

lbalmaceda:
This PR introduces the option to get a `Claim` as `Map<String, Object>`.

example usage:

```
DecodedJWT jwt = // decode or verify the jwt
Map<String, Object> userDataMap = jwt.getClaim("user-data").asMap();
```

Fixes https://github.com/auth0/java-jwt/issues/146

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/impl/JsonNodeClaim.java
+++ b/lib/src/main/java/com/auth0/jwt/impl/JsonNodeClaim.java
@@ -3,6 +3,7 @@ package com.auth0.jwt.impl;
 import com.auth0.jwt.exceptions.JWTDecodeException;
 import com.auth0.jwt.interfaces.Claim;
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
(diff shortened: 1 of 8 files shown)


lib/src/main/java/com/auth0/jwt/impl/JsonNodeClaim.java
```diff
@@ -91,6 +92,18 @@ public Date asDate() {
     }
 
     @Override
+    public Map<String, Object> asMap() throws JWTDecodeException {
+        ObjectMapper mapper = new ObjectMapper();
```
nikolaseu: not really this PR, but why isn't there just one `ObjectMapper` (even static) instead of creating a new one every time?
the same can be said about the `TypeReference`

lib/src/test/java/com/auth0/jwt/impl/JsonNodeClaimTest.java
```diff
@@ -207,6 +205,36 @@ public void shouldThrowIfListClassMismatch() throws Exception {
...
+        Claim claim = claimFromNode(value);
+
+        assertThat(claim, is(notNullValue()));
+        Map<String, Object> backMap = claim.asMap();
+        assertThat(backMap, is(notNullValue()));
+        assertThat(backMap, hasEntry("text", (Object) "extraValue"));
+        assertThat(backMap, hasEntry("number", (Object) 12));
+        assertThat(backMap, hasEntry("boolean", (Object) true));
+        assertThat(backMap, hasKey("object"));
+        assertThat((Map<String, Object>) backMap.get("object"), IsMapContaining.hasEntry("something", (Object) "else"));
```
nikolaseu: we need more "high level" tests here IMHO, we're just using the same implementation to encode & decode.
isn't somewhere tests that use a plain json string? I think we should add there one example claim of this type and test that everything works as expected

lib/src/main/java/com/auth0/jwt/impl/JsonNodeClaim.java
lbalmaceda: It's on our backlog to enhance this. We're thinking on moving the implementation to use Jackson's `Reader` and `Writer`.

lib/src/test/java/com/auth0/jwt/impl/JsonNodeClaimTest.java
lbalmaceda: I think I added the tests you're asking for, considering this PR adds only the ability to "get as map" and not the verify/create.

Issue #153, opened 2017-03-15
Add dedicated Exception for token expiration 


Spyna:
in class `com.auth0.jwt.JWTVerifier` , method `assertValidDateClaim`,  instead of throwing an `InvalidClaimException` it could be nice to throw a dedicated Exception, so that server and clients could better handle the token expiration.

PR #155, opened 2017-03-16
added date validation dedicated exception

Spyna:
for issue https://github.com/auth0/java-jwt/issues/153

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
@@ -1,15 +1,5 @@
 package com.auth0.jwt;
 
-import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
-import org.apache.commons.codec.binary.Base64;
-
 import com.auth0.jwt.algorithms.Algorithm;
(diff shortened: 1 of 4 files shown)


hzalaz:
Thanks for the PR @Spyna can you try to reduce the diff the PR has, e.g. avoid reordering the class elements since its hard to tell what you changed?

Spyna:
@hzalaz I' ve restored the original order. 
The modified method is `assertValidDateClaim`

lbalmaceda:
Hi @Spyna 
Thanks for the changes. We think it should throw a specific Exception _only_ in case the `exp` claim is invalid: lets call that `TokenExpiredException` and make it extend `JWTVerificationException`, as it's not that the claim it's invalid but rather that the token was ok but it has expired. Also let's keep the `InvalidClaimException` in case of invalid `nbf` and `iat` claims. Sounds good to you?

The changes should be done in the `assertValidDateClaim` method: When the boolean is true means we are here for the `exp` claim, and if it's invalid we should throw this new exception type. When the boolean is false, means it's one of the other two claims and if it's invalid we should throw the old exception. You can also split the method in 2 and handle these scenarios in a separate way if you want. Please make sure to add/update the tests to check the new exception.

Cheers

Spyna:
Sounds great to me, 
I'll modify the code and push, 

thank you

lbalmaceda:
@Spyna Please rollback the code formating changes in order to leave a clearer diff of the `JWTVerifier` class. Besides that it's looking fine. Thanks

PR #157, opened 2017-03-30
Add Sign/Verify of Long type claims

vrancic:
I added Long type for claims and updated tests.

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -9,7 +9,6 @@ import com.auth0.jwt.impl.PublicClaims;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationConfig;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import org.apache.commons.codec.binary.Base64;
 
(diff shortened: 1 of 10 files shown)


lbalmaceda:
Hi @vrancic 
Thanks for the PR 🎉 ! Before merging it there are a few tests missing I'll need you to add:
- `NullClaim`: Test that it returns a null Long. [(ref)](https://github.com/auth0/java-jwt/blob/master/lib/src/test/java/com/auth0/jwt/impl/NullClaimTest.java)
- `JWTVerifier`: Test that a Long claim can be set and verified [(ref)](https://github.com/auth0/java-jwt/blob/master/lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java#L214-L223) and test that a Long claim has a different value and the exception is raised [(ref)](https://github.com/auth0/java-jwt/blob/master/lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java#L147-L156). 

Check the coverage reports if you need more insight.
Cheers.

vrancic:
Hi @lbalmaceda 

All tests are passing now, so I think that this version is ok. If everything is ok, when can I expect the changes to be approved and library added to official maven repository?

Regards

lbalmaceda:
@vrancic Thanks for the update. Looks good! 
I'll be happy to merge the PR but for consistence I think it should also add a method to set a `Long` claim in the `JWTCreator.Builder`. Check [this](https://github.com/auth0/java-jwt/blob/master/lib/src/main/java/com/auth0/jwt/JWTCreator.java#L189) as an example and please add the corresponding tests.
Once approved it can take up to 2 weeks for us to publish it. 

Cheers

Issue #135, opened 2017-01-20
Cannot add map objects to claims


davide-maestroni:
I have troubles migrating from version 2 to version 3 of the library. In particular, there seems to be no way to generate a signature for an object like the one described [here](https://auth0.com/docs/api/management/v2), since the "scopes" claim is a map of objects. The `withClaim()` methods support only basic types and arrays of strings and integers, but I believe several more types (like maps) should be supported as well. Why not just exposing a method `withClaim(String name, Object value)`?

hzalaz:
@davide-maestroni we thought about that and we decided to keep the types of the claims simple since the jwt structure should be very simple and the bigger the object graph you have the bigger the token is. 
Also there is no way for us to ensure that the POJO can be serialized to JSON, we need to leverage that to the library user.

About the scopes for api v2, you can also use `scope` instead of `scopes` and each scope space separated with the following format `entity:action`.

Will evaluate later how we can add some support for Maps at least

davide-maestroni:
@hzalaz Ok, I see your point, I'll go for the `scope` version then. Thank you for the quick reply.

kwkhaw:
@davide-maestroni same here. We need a way to add Map to the claims.

PR #159, opened 2017-04-05
Add possibility to define JWT Claim data from any object

mohic:


lbalmaceda:
Hi @mohic 
Thanks for the PR. Unfortunately we can't ensure the proper JSON serialization of any class, and we are not willing to provide a JSON serializer API as that would require our library to tie to a specific JSON library. 

See related discussion: https://github.com/auth0/java-jwt/issues/135 and https://github.com/auth0/java-jwt/issues/146

Issue #163, opened 2017-04-11
How do you create a Claim that is a collection?


rgibanez:
I'm trying to create a Claim which will be called 'custom' and will be a collection of key/value pairs.

Example:

`"user_id": "123456",
  "custom": {
    "first_name": "John",
    "last_name": "Smith"
  }`

I can create the Claim: user_id, but am unsure of the best way to add the Claim which is a collection.

`String token = JWT.create().withClaim("user_id", "123456").sign(algorithm);`

Any advise would be great.

lbalmaceda:
@rgibanez what you want to create is claims with Maps. That's not available and won't be at least for now. Check issues like https://github.com/auth0/java-jwt/issues/135

rgibanez:
@lbalmaceda - thanks for the fast response - any rough time frame for this? Otherwise I'll have to look for another library :(

lbalmaceda:
@rgibanez Unfortunately there's no ETA as we have other priorities right now. But we're looking into a solution that can include at least simple map support.
Cheers.

mklnwt:
IMHO you should add the possibility to add a string as [RawValue](https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/util/RawValue.html).
That way we can serialize our object ourselves and still get a "pretty" output.

ATM if we add a string claim, it will be escaped and all quotation etc. starts with a backslash which is really annoying and can lead to problem with usage on consumer side.

PR #164, opened 2017-04-13
[WIP] Add Custom Map Claim

gustavoamigo:
I need to create a claim that is a map, so added a method the JWT Builder to add custom map. Example:
```
{
    "user": {
        "id": "u1234",
        "accountId": "a5678",
        "firstName": "Joe",
        "lastName": "Tester",
        "email": "joe.tester@example.com",
        "paymentProviderId": null
    }
}
```

## TODO as requested by peer review. 

* [ ] The method must add claims of type Map<String, Object> (not Map).
* [ ] The readme states which types of claim can be set and read. Add a clear warning message
explaining the above, telling that only primitives and collections are allowed and that custom classes are not guaranteed to be handled just fine and may throw.
* [ ] Add the same warning to the javadoc of the withClaim() method that receives the Map.
* [ ] Add enough tests to assert that primitives and collections are supported, but that custom classes will throw an error. Keep in mind that this should work in create and decode functions.
See https://docs.referralsaasquatch.com/topics/json-web-tokens/

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -290,6 +290,19 @@ public final class JWTCreator {
         }
 
         /**
+         * Add a custom Map Claim with the given items
+         * @param name the Claim's name.
+         * @param map  the Claim's value.
+         * @return this same Builder instance.
+         * @throws IllegalArgumentException IllegalArgumentException if the name is null.
+         */
+        public Builder withMapClaim(String name, Map map) throws IllegalArgumentException {
+            assertNonNull(name);
+            addClaim(name, map);
+            return this;
+        }
+
+        /**
          * Creates a new JWT and signs is with the given algorithm
          *
          * @param algorithm used to sign the JWT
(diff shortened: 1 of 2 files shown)


gustavoamigo:
@lbalmaceda is there a channel where I can discuss this PR with you guys?

gustavoamigo:
Hi, I created a fork and published in [bintray repository](https://bintray.com/gustavoamigo/maven/java-jwt) because I need this feature now. I am willing

lbalmaceda:
Hi @gustavoamigo 
The thing is that we can't ensure proper serialization/deserialization of the values in the Map because they are `Objects`. It will work for primitives and collections but it might not for other classes. The API will NOT be opened to custom serializers/deserializers as we don't want to force the user to use a specific JSON library, so a solution might be to limit the Map contents somehow. There are many closed issues regarding this. Check for example https://github.com/auth0/java-jwt/issues/163

We can accept an implementation if contains the following:

- [ ] The method must add claims of type `Map<String, Object>` (not `Map`).
- [ ] The readme states which types of claim can be set and read. Add a clear warning message explaining the above, telling that only primitives and collections are allowed and that custom classes are not guaranteed to be handled just fine and may throw. 
- [ ] Add the same warning to the javadoc of the `withClaim()` method that receives the Map.
- [ ] Add enough tests to assert that primitives and collections are supported, but that custom classes will throw an error. Keep in mind that this should work in create and decode functions.

Cheers.

lbalmaceda:
Closing this as it looks abandoned. Ping me if you need me to open it @gustavoamigo.

complanboy2:
Hi @lbalmaceda ,

Your review is appreciated, https://github.com/auth0/java-jwt/pull/289 which is opened for this request. Thank you.

Issue #131, opened 2017-01-06
java.lang.NoSuchMethodError for Base64.decodeBase64(String)


medavox:
Hi, while using your library I encountered the following runtime exception, which crashes my app:

```
E/AndroidRuntime: FATAL EXCEPTION: main
	 Process: com.xxxx.yyyy, PID: 17510
	 java.lang.NoSuchMethodError: No static method decodeBase64(Ljava/lang/String;)[B in class Lorg/apache/commons/codec/binary/Base64; or its super classes (declaration of 'org.apache.commons.codec.binary.Base64' appears in /system/framework/org.apache.http.legacy.boot.jar)
		 at com.xxxx.yyyy.network.RealNetworkController$2.onResponse(RealNetworkController.java:172)
		 at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run(ExecutorCallAdapterFactory.java:68)
		 at android.os.Handler.handleCallback(Handler.java:739)
		 at android.os.Handler.dispatchMessage(Handler.java:95)
		 at android.os.Looper.loop(Looper.java:158)
		 at android.app.ActivityThread.main(ActivityThread.java:7224)
		 at java.lang.reflect.Method.invoke(Native Method)
		 at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230)
		 at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120)
```

This is caused by your use of the Apache Commons Codec library, in [JWTDecoder.java:38](https://github.com/auth0/java-jwt/blob/master/lib/src/main/java/com/auth0/jwt/JWTDecoder.java#L38):

```
headerJson = StringUtils.newStringUtf8(Base64.decodeBase64(parts[0]));
            payloadJson = StringUtils.newStringUtf8(Base64.decodeBase64(parts[1]));
```

I have investigated the issue, and I believe it's down to an obscure problem with Android.

The Problem
=====

Until [recently](https://developer.android.com/about/versions/marshmallow/android-6.0-changes.html#behavior-apache-http-client), Android was bundled with Apache's [HttpClient](http://hc.apache.org/httpcomponents-client-4.5.x/android-port.html), which depended on Commons Codec.

However, the version originally bundled with Android (1.3) *does not contain* the static `decodeBase64(String)` method. Furthermore, this ancient version overrides the version you specify in your lib's `build.gradle`.

[...]

Issue #191, opened 2017-08-03
Android Studio NoSuchMethodError


woliurain:
java.lang.NoSuchMethodError: No static method encodeBase64URLSafeString([B)Ljava/lang/String; in class Lorg/apache/commons/codec/binary/Base64; or its super classes (declaration of 'org.apache.commons.codec.binary.Base64' appears in /system/framework/org.apache.http.legacy.boot.jar)
                                                                    at com.auth0.jwt.JWTCreator.sign(JWTCreator.java:330)

lbalmaceda:
For Android if you just need to decode a JWT then you can use [this one](https://github.com/auth0/JWTDecode.Android) as the readme says.

Issue #185, opened 2017-06-28
 No static method decodeBase64


Kolyall:
I'm trying to run the lib on Android
My code:
 ```
String token =
"eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJfaWQiOiI1OTUzNjFjNDU0NGU2NDYxOTMxYWY1MzMiLCJhY2Nlc3MiOiJhdXRoIiwiaWF0IjoxNDk4NjQwNDA0fQ.dHlPdvq8-0h8SJeSZjVcJcW1WOW72HAL4oAb_cpzpsk"
 DecodedJWT jwt = JWT.decode(token);
 ```
Error:
```
Caused by: java.lang.NoSuchMethodError: No static method decodeBase64(Ljava/lang/String;)[B in class Lorg/apache/commons/codec/binary/Base64; or its super classes (declaration of 'org.apache.commons.codec.binary.Base64' appears in /system/framework/org.apache.http.legacy.boot.jar)
                                                                            at com.auth0.jwt.JWTDecoder.<init>(JWTDecoder.java:32)
                                                                            at com.auth0.jwt.JWT.decode(JWT.java:21)
```

lbalmaceda:
Apache commons [decodeBase64](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html#decodeBase64(java.lang.String)) isn't prepared for Android. The 3rd line of the [Readme](https://github.com/auth0/java-jwt) points Android users here -> https://github.com/auth0/JWTDecode.Android

PR #193, opened 2017-08-10
Fix bug when sign JWT on Android platform

laurentyhuel:
Fix bug https://github.com/auth0/java-jwt/issues/131 , https://github.com/auth0/java-jwt/issues/191, https://github.com/auth0/java-jwt/issues/185 : cannot sign JWT on Android platform.
I choose reflection to check if code is execute on Android platform why:
1) Because not need to embed a new library for instance:
dependencies {
    compileOnly "com.google.android:android:4.1.1.4"
}
2) Because it seems the more secure way to check if code is execute on Android platform
It' also possible to get platform with System.getProperty() : https://stackoverflow.com/questions/4519556/how-to-determine-if-my-app-is-running-on-android
but result could differ depends on os version, manufacturer, ...


lbalmaceda:
Ughh! 
Thanks for the PR. Personally, I don't like the use of reflection as it's too hacky and may break without us noticing. In addition, I don't think those 3 issues you mention are "bugs" because we decided up-front that this library was meant to be used on Java-only environments, leaving a side Android which has its own **decode** library [here](https://github.com/auth0/JWTDecode.Android). As mentioned in one of the issues, there's no need for you to sign or even verify a token from the Android client side. You *should trust* your server and its implementation. What's your use case?

laurentyhuel:
Our use case:
On android device, we receive a oauth token from few 3rd party OAuth servers (partners).
Android create a JWT with this oauth token and other properties.
Android call our server with JWT.
Our server get oauth otken from JWT.
Our server check oauth token to a 3rd party OAuth servers (partners).

So yes, you need to sign JWT on Android (it's not just for the fun).

I'm agree with you reflection, is not sexy, but in this case it seems to be the more robust way to do it.

I could add unit tests to coverage this part.

lbalmaceda:
There's no way to keep the secret/private_key secure in Android (anyone can see/modify it) so what's the point in signing the JWT? Ok, you could make the server use the public_key (in the case of HMAC) to verify the incoming signature, but even then anyone can sign tokens with your private_key as you're keeping it on the device, meaning your server would trust anything blindfolded. The tokens must always be signed on the server.
If you just want to use a JWT to wrap all the third-party tokens then that's just a JSON, the same as doing a post with a JSON body to your server with each token in a property or even an array of tokens.

PR #202, opened 2017-09-18
2017 Sep version updates

jspetrak:


lbalmaceda:
Josef, thanks for the contribution. I don't want to update gradle plugin just now.. But I've updated the rest of the dependencies in this https://github.com/auth0/java-jwt/pull/212.

PR #204, opened 2017-09-19
Method for adding strings as RawValue

mklnwt:
You should add the possibility to add a string as [RawValue](https://fasterxml.github.io/jackson-databind/javadoc/2.6/com/fasterxml/jackson/databind/util/RawValue.html).
That way we can serialize our objects ourselves and still get a "pretty" output.

ATM if we add a string claim, it will be escaped and all quotation etc. starts with a backslash which is really annoying and can lead to problem with usage on consumer side.

This could also help #163.

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -10,6 +10,8 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.util.RawValue;
+
 import org.apache.commons.codec.binary.Base64;
 
 import java.nio.charset.StandardCharsets;
@@ -287,6 +289,20 @@ public final class JWTCreator {
         public Builder withArrayClaim(String name, Long[] items) throws IllegalArgumentException {
             assertNonNull(name);
             addClaim(name, items);
+            return this;
+        }
+
+        /**
+         * Add a custom Claim value as {@link RawValue}. It will not be escaped to a string this way. 
+         *
+         * @param name  the Claim's name.
+         * @param value the Claim's value.
+         * @return this same Builder instance.
+         * @throws IllegalArgumentException if the name is null.
+         */
+        public Builder withRawClaim(String name, String value) throws IllegalArgumentException {
+            assertNonNull(name);
+            addClaim(name, new RawValue(value));
             return this;
         }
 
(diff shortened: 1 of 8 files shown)


lbalmaceda:
Yes, this might work while the `Map<String, Object>` support is missing. How would you get them back though, does the `Claim.asString` work as expected?

mklnwt:
Good point. New method `asRawValue` can now be used.

This has an additional benefit. While you can use `as(Class<T> tClazz)` to get custom POJOs (or DTOs), you heavily rely on the deserializer used by this library. With the new method you can use your own deserializer.

That is also kind of the `Map<String, Object>` support. You serialize your Map or POJO in advance and put it into `withRawClaim`.

lbalmaceda:
I'm having second thoughts. Honestly, I would rather have a good implementation using `Map<String, Object>` (like the PR someone was making) than the raw way. We should be providing a good public API within the restriction of not using specific implementations like Jackson or Gson. I'll close this PR and see if in the next 2 weeks I can give it some time to add that method. Thanks for the interest.

mklnwt:
I would love to see both.

As I mentioned, the raw way will provide you the flexibility to handle the payload how ever you like.
Even with `Map<String, Object>` you would probably end up also needing a Method for just `Object`.
And even then you will always kind of rely on the implementation as you often serialize based on annotations.

mklnwt:
In fact I would run into a problem without a raw possibility.

We are using JBoss EAP 6 which includes Jackson 1.9.9, the API totally differs from Jackson 2.x (org.codehaus.jackson instead of com.fasterxml.jackson and different annotations).
It would mean a lot more effort to convert an object to `Map<String, Object>` or alike than to just use our existing `org.codehaus.jackson.map.ObjectMapper` and pass the string as raw value.

Using `withClaim(String name, String value)` is a workaround but not really clean as we end up with an escaped string (e.g. `{... "custom-claim": "{\"name\":\"George\",\"id\":1}" ...}`).

mklnwt:
Any update on this?

lbalmaceda:
No, this is not a priority for us right now and we don't want to have "raw string" support. I'd accept a `Map<String, Object>` implementation though, like I said before.

lbalmaceda:
IMHO there's no need to nest objects inside the JWT. Payload should be plain, as simple as possible.

mklnwt:
Would it be possible to make the JWTCreator / Builder extensible so that we can provide our own payload type? Maybe with a callback function?

IMHO a JWT should be nestable with all of JSONs features. E.g. you could use it as a session mechanism with user roles, settings, etc. With a flat structure (without a least a few levels) this will become confusing.

lbalmaceda:
Tokens are normally sent on every http request. If you add a huge payload, the size increases and the requests take longer. I understand your need but can't think of a nice solution that doesn't involve using raw string nor using Jackson/Gson APIs in our public API. I can't accept those. If you think there's one please share a gist with your thoughts and I'll review it with my team.

This is not a solution but a workaround, you could use custom claims with namespaces like `https://app.company.com/user/id` and `https://app.company.com/user/email` just to name something, and have some method that reassembles that into the POJO you need.

mklnwt:
That's a good idea. I'm not sure if this would save me any space though.

mklnwt:
Can you declare addClaim() in JWTCreator as protected so I can override it myself in our project.
I can understand your intention to not add the RawValue to the project, but I would love the library to be at least extensible.

PR #206, opened 2017-09-30
enhance the library to include the ability to add headers, and retrieve all headers on parse

DinoChiesa:


lbalmaceda:
@DinoChiesa why do you need to iterate through the list of header claims? Typically you already know what to look for after parsing the header, and there's already a method to obtain one header claim by name. That said, the PR also contains lots of changes that are not only unnecessary but also not part of the intent of this PR (what the title describes). I know its being some time since you submitted this. If you're still interest let's discuss about the header claims.

PR #212, opened 2017-10-23
Rework ECDSA

lbalmaceda:
This PR fixes a lot of ECDSA Algorithm issues. The problem was mainly that the signature format conversion was being made on one way but not in the other. I've added some tests that mimic the signature format and test the converter methods. What I'm not so sure about is if we should have a different exception message when the total signature length or the R/S numbers length is not the expected. We could throw the same error message "invalid signature length" for all 3 cases. Check lines [104](https://github.com/auth0/java-jwt/compare/rework-ecdsa?expand=1#diff-535ab2373cdedf6aa0c947591a23dddeR104), [113](https://github.com/auth0/java-jwt/compare/rework-ecdsa?expand=1#diff-535ab2373cdedf6aa0c947591a23dddeR113) and [125](https://github.com/auth0/java-jwt/compare/rework-ecdsa?expand=1#diff-535ab2373cdedf6aa0c947591a23dddeR125).

Tests should run against both JDK 7 & JDK 8. I don't know why but tests were failing on JDK8, something we didn't caught because we never run the tests on that JDK. I added a separate PR to use _CircleCI Workflows_ to attempt to run the tests in both JDKs, but because the tests don't pass in master merging that to master will "break master" until this PR is merged. I'd add it sourced from this last commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,7 @@ Temporary Items
 
 # IntelliJ
 /out/
+/lib/out/
 
 # mpeltonen/sbt-idea plugin
 .idea_modules/
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -34,13 +34,13 @@ compileJava {
 }
 
 dependencies {
-    compile 'com.fasterxml.jackson.core:jackson-databind:2.8.4'
-    compile 'commons-codec:commons-codec:1.10'
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.9.2'
+    compile 'commons-codec:commons-codec:1.11'
     testCompile 'org.bouncycastle:bcprov-jdk15on:1.58'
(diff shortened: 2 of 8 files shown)


lbalmaceda:
Yes. The `ECDSABouncyCastleProviderTests` is a copy of the `ECDSAAlgorithmTests` with a BeforeClass and AfterClass that setup the preferred provider. Because we don't hardcode the provider name when requiring one in our logic, it will always pick the preferred one.. which in one test file is the "JDK's default" and in the other one is "bc".

I've added a small test to ensure that the first provider is actually BC.
https://github.com/auth0/java-jwt/pull/212/files#diff-d52011bdc8b4ab49a8ed80b686f1978bR70

I've removed invalid tests as DER signature encoding should NOT be allowed in a JWT and added the corresponding ones to check that it fails on those cases.

PR #231, opened 2018-01-16
Changes for java6 compatibility

hgastaud:
Guys this are the changes for making the code work with Java 1.6

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -19,355 +19,313 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 
 /**
- * The JWTCreator class holds the sign method to generate a complete JWT (with
- * Signature) from a given Header and Payload content.
+ * The JWTCreator class holds the sign method to generate a complete JWT (with Signature) from a given Header and Payload content.
  */
 @SuppressWarnings("WeakerAccess")
 public final class JWTCreator {
 
-	private final Algorithm algorithm;
-	private final String headerJson;
-	private final String payloadJson;
+    private final Algorithm algorithm;
+    private final String headerJson;
+    private final String payloadJson;
 
-	private JWTCreator(Algorithm algorithm, Map<String, Object> headerClaims, Map<String, Object> payloadClaims) throws JWTCreationException {
-		this.algorithm = algorithm;
-		try {
-			ObjectMapper mapper = new ObjectMapper();
-			SimpleModule module = new SimpleModule();
-			module.addSerializer(ClaimsHolder.class, new PayloadSerializer());
-			mapper.registerModule(module);
-			mapper.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true);
-			headerJson = mapper.writeValueAsString(headerClaims);
-			payloadJson = mapper.writeValueAsString(new ClaimsHolder(payloadClaims));
-		} catch (JsonProcessingException e) {
-			throw new JWTCreationException("Some of the Claims couldn't be converted to a valid JSON format.", e);
-		}
-	}
+    private JWTCreator(Algorithm algorithm, Map<String, Object> headerClaims, Map<String, Object> payloadClaims) throws JWTCreationException {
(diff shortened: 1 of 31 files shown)


lbalmaceda:
Hernan. The diff is full of code format changes or imports being re-ordered. Can you please fix them so I can find the actual code changes?

image

hgastaud:
Sorry for the code format, I've already made the rollback for those format changes.

Regards,
Hernan.

lbalmaceda:
It looks much better now but there are still some lines left. Please check the [files](https://github.com/auth0/java-jwt/pull/231/files) section of this PR to find them. Note that this project uses whitespaces instead of tabs, and that if the _imports_ haven't changed they shouldn't be reordered. Lastly, it's OK to replace `some.class.*` with `some.class.a` and `some.class.b`. 👍

hgastaud:
Luciano,

I've modified all the issues you'd mentioned in your last comment, can you check it?

Regards,
Hernan.

lib/src/main/java/com/auth0/jwt/JWTCreator.java
```diff
@@ -327,11 +327,11 @@ private void addClaim(String name, Object value) {
     }
 
     private String sign() throws SignatureGenerationException {
-        String header = Base64.encodeBase64URLSafeString(headerJson.getBytes(StandardCharsets.UTF_8));
-        String payload = Base64.encodeBase64URLSafeString(payloadJson.getBytes(StandardCharsets.UTF_8));
+        String header = Base64.encodeBase64URLSafeString(headerJson.getBytes(Charset.forName("UTF-8")));
```
lbalmaceda: Create an interface under the `com.auth0.jwt.interfaces` package to keep a constant and reference it everywhere instead of hardcoding the string value.

lib/src/main/java/com/auth0/jwt/algorithms/ECDSAAlgorithm.java
```diff
@@ -49,7 +47,7 @@ public void verify(DecodedJWT jwt) throws SignatureVerificationException {
             if (!valid) {
                 throw new SignatureVerificationException(this);
             }
-        } catch (NoSuchAlgorithmException | SignatureException | InvalidKeyException | IllegalStateException e) {
+        } catch (Exception e) {
```
lbalmaceda: expand the catch clause instead of capturing a generic `Exception`. Capture each of them and make all do the same. (Throw a wrapped exception in this particular case)

hgastaud:
Luciano,

I made the changes that you've request in your previous comment.
About the testing issues:
1) For compilation with Java 1.6 I had to make a downgrade for two dependencies
2) Tests that runs with EC Algorithm wont work because this algorithm appears in Java 1.7.

So, perhaps merging this PR brings you the risk having only the 66% of the passed tests on your framework. I won't be able to add this algorithm to your code because it will require some time, which I don't have it right now. 

Regards,
Hernan.

lbalmaceda:
The jackson-databind downgrade looks good as the methods we use haven't changed, and the other one really doesn't bother me as it's just for testing. It's true. I've tried to run the tests and almost all of the ECDSA ones fail because of what you say. I read that probably with BouncyCastle it could be achieved, but it's not currently in our plans to change the crypto provider. I've also found a jdk6 docker image so it could be possible to add the CI build, but if it always fail because of that algorithm it won't do much good to show that result on the repo. 

Anyway, thanks for the contribution. I'm looking to merge this on the next week. I'll leave a few tasks below for me to do before that:
- [ ] Add a disclaimer on the README file saying we partially support 1.6 for all algorithms but Elliptic Curve, and that although it was tested once it's not currently being tested on the CI builds.
- [ ] Add the same disclaimer on ECDSA algorithm classes or public methods javadocs.

lbalmaceda:
Heads up 👋 . I'm holding off this PR due to a recently discovered security issue still not fixed in one of the JDK 6 dependency downgrades.

obecker:
You might consider introducing your own constant for `Charset.forName("UTF-8)` to prevent doing the lookup again and again.

lbalmaceda:
@hgastaud I'm closing this one. Downgrading the dependencies introduces security concerns we can't deal with, and no alternate version is present compatible with JDK6 to fill this. Thanks a lot for your time and contribution. 👍

PR #239, opened 2018-02-21
Token with custom objects as claims

EL-Greco-Domi:
This pull requests adds the ability to create tokens with custom objects as claims. 

## Motivation
It was possible to verify tokens with JSON objects as claim values, but it was not possible to create them. The public API didn't provide the necessary methods to add custom objects to the token as a claim value. 
This pull request enables a user of this library to do so, furthermore it is now also possible to pass mandatory object claims during verification of the token.

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -186,7 +186,7 @@ public final class JWTCreator {
          * @return this same Builder instance.
          * @throws IllegalArgumentException if the name is null.
          */
-        public Builder withArrayClaim(String name, Object... items) throws IllegalArgumentException {
+        public <T> Builder withArrayClaim(String name, @SuppressWarnings("unchecked") T... items) throws IllegalArgumentException {
             assertNonNull(name);
             addClaim(name, items);
             return this;
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -15,6 +15,7 @@ import org.apache.commons.codec.binary.Base64;
 import java.nio.charset.StandardCharsets;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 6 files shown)


jeffsteinmetz:
It would be beneficial to have this, there has been no work around for quite some time now.

jeffsteinmetz:
Thank you @EL-Greco-Domi , any ideas on why Java 7 is failing at (but Java 8 is ok).  This MR would be handy!

```
com.auth0.jwt.JWTCreatorTest > shouldAcceptCustomClaimOfTypeObject FAILED
    java.lang.AssertionError at JWTCreatorTest.java:339
```

jeffsteinmetz:
This merge would be very handy, thank you @EL-Greco-Domi 

Any thoughts why this fails on Java 7, but not 8?

```
com.auth0.jwt.JWTCreatorTest > shouldAcceptCustomClaimOfTypeObject FAILED
    java.lang.AssertionError at JWTCreatorTest.java:339
```

lbalmaceda:
Please check the open/closed issues regarding this same matter. This is definitely something we don't want to accept in a generic way. If you still want to try, an implementation we suggested back then is detailed [here](https://github.com/auth0/java-jwt/pull/164#issuecomment-295764851).

EL-Greco-Domi:
Hi,
sorry I was quite busy. 

Regarding your concerns in the linked issue:
I do understand that we can't guarantee that the custom objects will be serialized in a way the user wants, but it could be that the default handling of the ObjectMapper is excatly what the user wants, and limiting his choice just because we could handle some Objects wrong is IMHO just too restrictive. 
I also understand that you don't want to force a specific library onto users, but for the sake of usability it would be really handy to be able to register custom serializers, either as a method on one of the library classes or using the @JsonSerialize-Annotation.
Of course all of the above could be achieved by your suggestion of allowing Maps with the standard Java-Types, but that leads to just one result, the user is forced to implement the JSON parsing of custom objects by mapping his object's members into the according java types and putting them in a Map at the right position, although his custom object would be handled by ObjectMapper just fine.

If you really really don't wont to open up the library, there still is the possibility to create an own interface which is about the same like the one of jackson (StdSerializer), so just like this:

```java
public interface ObjectSerializer<T> {

	public String serialize(T item);
	
	public Class<T> getClassToHandle();
}

```

A user can pass custom objects, if he's unhappy with the result he can register custom serializers which extend this interface. we can then wrap his custom serializer inside a StdSerializer and pass it to the ObjectMapper as serializer for the specified class. Using this way we don't force a specific library onto the users and still have the benefits of being able to use custom objects.

Cheers,
Dominic

Issue #240, opened 2018-02-21
Should throw JWTDecodeException in case date based claims like "exp" are specified in the wrong format. 


uwinkelvos:
We saw an issue, that when accidentally specifying an "exp" claim as iso string the exp claim is silently ignored. Imho a  JWTDecodeException should be thrown.

See: 
https://github.com/auth0/java-jwt/blob/7dcea427df1db2b948911e70aef2e9d7f0b7cfad/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java#L68

lbalmaceda:
Yes, that makes sense. 👍

PR #241, opened 2018-02-21
Throw JWTDecodeException when date claim format is invalid

lbalmaceda:
WDYT about the exception message?

Fixes https://github.com/auth0/java-jwt/issues/240

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
+++ b/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
@@ -69,7 +69,7 @@ class PayloadDeserializer extends StdDeserializer<Payload> {
             return null;
         }
         if (!node.canConvertToLong()) {
-            throw new JWTDecodeException(String.format("The claim '%s' contained an unexpected value.", claimName));
+            throw new JWTDecodeException(String.format("The claim '%s' contained a non-numeric date value.", claimName));
         }
         final long ms = node.asLong() * 1000;
         return new Date(ms);
--- a/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
+++ b/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
@@ -65,8 +65,11 @@ class PayloadDeserializer extends StdDeserializer<Payload> {
 
     Date getDateFromSeconds(Map<String, JsonNode> tree, String claimName) {
         JsonNode node = tree.get(claimName);
-        if (node == null || node.isNull() || !node.canConvertToLong()) {
+        if (node == null || node.isNull()) {
             return null;
+        }
+        if (!node.canConvertToLong()) {
+            throw new JWTDecodeException(String.format("The claim '%s' contained an unexpected value.", claimName));
         }
         final long ms = node.asLong() * 1000;
         return new Date(ms);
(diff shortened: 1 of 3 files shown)


lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
```diff
@@ -65,9 +65,12 @@ public Payload deserialize(JsonParser p, DeserializationContext ctxt) throws IOE
 
     Date getDateFromSeconds(Map<String, JsonNode> tree, String claimName) {
         JsonNode node = tree.get(claimName);
-        if (node == null || node.isNull() || !node.canConvertToLong()) {
+        if (node == null || node.isNull()) {
             return null;
         }
+        if (!node.canConvertToLong()) {
+            throw new JWTDecodeException(String.format("The claim '%s' contained an unexpected value.", claimName));
```
uwinkelvos: or maybe better non-numeric

lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
```diff
@@ -198,13 +198,15 @@ public void shouldGetNullDateWhenParsingNull() throws Exception {
...
+        exception.expect(JWTDecodeException.class);
+        exception.expectMessage("The claim 'key' contained an unexpected value.");
+
         Map<String, JsonNode> tree = new HashMap<>();
         TextNode node = new TextNode("123456789");
         tree.put("key", node);
 
-        Date date = deserializer.getDateFromSeconds(tree, "key");
-        assertThat(date, is(nullValue()));
+        deserializer.getDateFromSeconds(tree, "key");
```
cocojoe: So no assertions as now caught by the exception.expect? if `getDateFromSeconds` explodes?

lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
```diff
@@ -65,9 +65,12 @@ public Payload deserialize(JsonParser p, DeserializationContext ctxt) throws IOE
 
     Date getDateFromSeconds(Map<String, JsonNode> tree, String claimName) {
         JsonNode node = tree.get(claimName);
-        if (node == null || node.isNull() || !node.canConvertToLong()) {
+        if (node == null || node.isNull()) {
```
cocojoe: So non Java person here, what is the difference here in checking for null?

lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
lbalmaceda: Exactly. Because the ExpectedException is "none" for every method but for those in which we override it, when we run tests they expect (or don't) to fail with X exception and Y message.

lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
lbalmaceda: `node` is the instance of the `JsonNode` we're checking. The check is asserting that the instance of this object is `null` OR that the parsed JSON value is null. This second check wouldn't be possible if the instance is `null` as that would throw a NPE when calling some method on it.

Issue #236, opened 2018-01-30
Refactor HMACAlgorithm#getSecretBytes() so that it does not throw UnsupportedEncodingException


dmak:
It is easily possible to modify [`HMACAlgorithm#getSecretBytes()`](https://github.com/auth0/java-jwt/blob/0001be066d0e6547792480de0dac099a649a6b02/lib/src/main/java/com/auth0/jwt/algorithms/HMACAlgorithm.java#L42) so that it does not throw `UnsupportedEncodingException`:
* Use `java.nio.charset.StandardCharsets.UTF_8` from Java7.
* Use `org.apache.commons.codec.Charsets.UTF_8` from Commons Codec.

Example:

    return secret.getBytes(Charsets.UTF_8);

PR #242, opened 2018-02-22
Fix for issue #236 - refactored HMACAlgorithm so that it doesn't throw an UnsupportedEncodingException

obecker:
- replaced call of String.getBytes(String charsetName) with String.getBytes(Charset charset)


skjolber:
@lbalmaceda bump

PR #244, opened 2018-02-26
Bump Jackson dependency

skjolber:
Bump dependencies to fix a [security bug](https://nvd.nist.gov/vuln/detail/CVE-2017-17485) in Jackson.


skjolber:
@lbalmaceda bump

lbalmaceda:
Thanks. Please remove the owasp plugin as we're already using another service.

PR #248, opened 2018-03-13
getting HMAC* Algorithm does not throw UnsupportedEncodingException

yagee-de:
using StandardCharsets.UTF_8 instead of CharEncoding.UTF_8 does not
require to handle CharEncoding.UTF_8


yagee-de:
double post: https://github.com/auth0/java-jwt/pull/242 is not merged yet

PR #194, opened 2017-08-11
Using ObjectReader to cache Jackson bindings when parsing JWT

jack-harrison:
Hey Guys,

While perf testing a web service which uses auth0-spring-security-api, I noticed that a chunk of performance was getting taken up by Jackson's ObjectMapper.readValue due to the reflection required on every call to set up its bindings.

Thankfully, Jackson provides an ObjectReader which caches the bindings for us, and is completely thread safe once after initial configuration. Since all configuration happens in the JWTParser header, this is perfectly safe to do and provides some significant performance improvements.

Here is a quick and dirty test I wrote to test the performance:
```java
public class PerfTest {

    private static final String TOKEN = "eyJhbGciOiJIUzI1NiIsImN0eSI6IkpXVCJ9.eyJpc3MiOiJhdXRoMCJ9.mZ0m_N1J4PgeqWmi903JuUoDRZDBPB7HwkS4nVyWH1M";
    private static final int COUNT = 10000;
    private static final int THREADS = 50;

    @BeforeClass
    public static void warmup() {
        JWT jwt = new JWT();
        for (int i = 0; i < 1000; i++) {
            jwt.decodeJwt(TOKEN);
        }

        for (int i = 0; i < 1000; i++) {
            JWT.decode(TOKEN);
        }
    }

    @Test
    public void singleInstancePerformance() throws Exception {
        long start = System.nanoTime();

        JWT jwt = new JWT();
        runSingleThread(() -> jwt.decodeJwt(TOKEN));

        System.out.println("singleInstancePerformance:" + (System.nanoTime() - start) / 1000000L);
    }

    @Test
    public void singleInstanceThreadedPerformance() throws Exception {
        long start = System.nanoTime();

        JWT jwt = new JWT();
        runThreaded(() -> jwt.decodeJwt(TOKEN));

        System.out.println("singleInstanceThreadedPerformance:" + (System.nanoTime() - start) / 1000000L);
    }

    @Test
    public void staticPerformance() throws Exception {
        long start = System.nanoTime();

        runSingleThread(() -> JWT.decode(TOKEN));

        System.out.println("staticPerformance:" + (System.nanoTime() - start) / 1000000L);
    }

    @Test
    public void staticThreadedPerformance() throws Exception {
        long start = System.nanoTime();

        runThreaded(() -> JWT.decode(TOKEN));

        System.out.println("staticThreadedPerformance:" + (System.nanoTime() - start) / 1000000L);
    }

[...]

PR #255, opened 2018-05-30
Performance improvements

skjolber:
Changes from https://github.com/auth0/java-jwt/pull/194 plus various minor improvements. 
 * crypto helper is more application-specific to JWT in taking in header and payload as different arguments.  
 * passing ObjectReader around where it might be used
 * directly generate json in PayloadSerializer
 * default HashMap sizes


skjolber:
@lbalmaceda any hope for merging this? Also the build fail on java 7 due to some connection reset and code coverage seems to be not working either(?).

lib/src/main/java/com/auth0/jwt/impl/JWTParser.java
```diff
@@ -47,19 +67,11 @@ static ObjectMapper getDefaultObjectMapper() {
...
-    <T> T convertFromJSON(String json, Class<T> tClazz) throws JWTDecodeException {
-        if (json == null) {
-            throw exceptionForInvalidJson(null);
-        }
-        try {
-            return mapper.readValue(json, tClazz);
-        } catch (IOException e) {
-            throw exceptionForInvalidJson(json);
-        }
+    private static JWTDecodeException decodeException() {
```
lbalmaceda: there should be no reason to make this 2 methods `static`. Does this have any impact on the performance?

lib/src/main/java/com/auth0/jwt/impl/PayloadImpl.java
```diff
@@ -73,14 +76,14 @@ public String getId() {
...
     @Override
     public Claim getClaim(String name) {
-        return extractClaim(name, tree);
+        return extractClaim(name, tree, objectReader);
     }
 
     @Override
     public Map<String, Claim> getClaims() {
-        Map<String, Claim> claims = new HashMap<>();
+        Map<String, Claim> claims = new HashMap<>(tree.size() * 2);
```
lbalmaceda: What's the rationale to pick this default value?

lib/src/main/java/com/auth0/jwt/impl/PayloadSerializer.java
```diff
@@ -21,37 +20,47 @@ private PayloadSerializer(Class<ClaimsHolder> t) {
...
                 case PublicClaims.EXPIRES_AT:
                 case PublicClaims.ISSUED_AT:
                 case PublicClaims.NOT_BEFORE:
-                    safePayload.put(e.getKey(), dateToSeconds((Date) e.getValue()));
+                    gen.writeFieldName(e.getKey());
+                    gen.writeNumber(dateToSeconds((Date) e.getValue()));
                     break;
                 default:
+                    gen.writeFieldName(e.getKey());
                     if (e.getValue() instanceof Date) {
```
lbalmaceda: this `default` case could be merged with the 3 known claims above (which are dates as well), or rephrasing, this could be deleted:

```java
case PublicClaims.EXPIRES_AT:
case PublicClaims.ISSUED_AT:
case PublicClaims.NOT_BEFORE:
   gen.writeFieldName(e.getKey());
   gen.writeNumber(dateToSeconds((Date) e.getValue()));
   break;
```

lib/src/main/java/com/auth0/jwt/algorithms/CryptoHelper.java
```diff
@@ -2,31 +2,50 @@
...
         final Mac mac = Mac.getInstance(algorithm);
         mac.init(new SecretKeySpec(secretBytes, algorithm));
-        return mac.doFinal(contentBytes);
+        mac.update(headerBytes);
+        mac.update(DOT);
+        return mac.doFinal(payloadBytes);
     }
 
-    boolean verifySignatureFor(String algorithm, PublicKey publicKey, byte[] contentBytes, byte[] signatureBytes) throws NoSuchAlgorithmException, InvalidKeyException, SignatureException {
+    boolean verifySignatureFor(String algorithm, PublicKey publicKey, String header, String payload, byte[] signatureBytes)  throws NoSuchAlgorithmException, InvalidKeyException, SignatureException {
```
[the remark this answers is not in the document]
skjolber: Well, the header/payload from DecodedJWT are strings, so it is good to pass them directly so to the helper. Also there is quite a lot of unit test calls to this convenice method. So I guess keeping them around is fair.

lib/src/main/java/com/auth0/jwt/algorithms/CryptoHelper.java
```diff
@@ -2,31 +2,50 @@
...
+
 import java.security.*;
 
 class CryptoHelper {
+	
+	private static final byte DOT = (byte)46;
 
-    boolean verifySignatureFor(String algorithm, byte[] secretBytes, byte[] contentBytes, byte[] signatureBytes) throws NoSuchAlgorithmException, InvalidKeyException {
-        return MessageDigest.isEqual(createSignatureFor(algorithm, secretBytes, contentBytes), signatureBytes);
+    boolean verifySignatureFor(String algorithm, byte[] secretBytes, String header, String payload, byte[] signatureBytes) throws NoSuchAlgorithmException, InvalidKeyException {
```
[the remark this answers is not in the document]
skjolber: See below comment

lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
```diff
@@ -364,9 +364,11 @@ public String toString() {
...
      * Sign the given content using this Algorithm instance.
      *
-     * @param contentBytes an array of bytes representing the base64 encoded content to be verified against the signature.
+     * @param headerBytes an array of bytes representing the base64 encoded header content to be verified against the signature.
+     * @param payloadBytes an array of bytes representing the base64 encoded payload content to be verified against the signature.
      * @return the signature in a base64 encoded array of bytes
      * @throws SignatureGenerationException if the Key is invalid.
      */
-    public abstract byte[] sign(byte[] contentBytes) throws SignatureGenerationException;
+    public abstract byte[] sign(byte[] headerBytes, byte[] payloadBytes) throws SignatureGenerationException;
```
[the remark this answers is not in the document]
skjolber: Yes. Basically this narrows the use for Algorithm to just signing tokens, which is in line with its own class description (in the javadoc).

lib/src/main/java/com/auth0/jwt/impl/PayloadImpl.java
skjolber: Crude estimate of necessary capacity to avoid rehashing and collision. 

https://stackoverflow.com/questions/10901752/what-is-the-significance-of-load-factor-in-hashmap

lib/src/main/java/com/auth0/jwt/impl/PayloadSerializer.java
skjolber: Deleted according to your comment.

skjolber:
@lbalmaceda see updates / comments, sorry for slow reaction on QA.

lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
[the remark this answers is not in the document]
lbalmaceda: I don't disagree that this is an improvement. But it will break customers who are implementing their own Algorithms.. I can't see many customers doing that but some feature requests have been opened in the past asking us to extend the algorithms we support. Let me discuss this internally. Maybe with a good readme note and proper changelog notes we can accept it. I can't promise anything.

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -51,7 +54,7 @@ static Verification init(Algorithm algorithm) throws IllegalArgumentException {
             }
 
             this.algorithm = algorithm;
-            this.claims = new HashMap<>();
+            this.claims = new HashMap<>(48);
```
lbalmaceda: No magic numbers please. Can we move it into a constant and give it a proper name/description?

lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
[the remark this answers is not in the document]
lbalmaceda: Re above: No breaking change allowed (and no major plans for the time being). We need to find an alternative to this. Normally I'd say call the new method from the old one and deprecate it, but in this case there's no way for you to figure out at what point of the `contentBytes` array is the half of it in order to separate it into header and payload bytes. What if you add the new sign signature and keep both methods side-to-side? I guess the new one can call the old one by merging both header and payload bytes with a "." in-between, so not to repeat the sign method's body code. If not you will have to rollback the changes for this class.

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
skjolber: On second thought, 16 (the default) is probably okey here.

lbalmaceda: Agree. I've never seen a token with that many claims 😆

lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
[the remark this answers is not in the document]
skjolber: I've added a default implementation of the new method (sign with header and payload) in Algorithm, but overridden that in the known subclasses. This should ensure backwards compatibility for those implementing their own algorithms, at the same time as speeding things up for the rest of us. 

CryptoHelper has now both the new and the old methods, but they're all quite small.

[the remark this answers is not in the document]
skjolber: @lbalmaceda If this solution is acceptable, I'll add a few of the old unit tests for code coverage (see failing build checks).

lib/src/main/java/com/auth0/jwt/algorithms/RSAAlgorithm.java
```diff
@@ -64,6 +63,19 @@ public void verify(DecodedJWT jwt) throws SignatureVerificationException {
         }
     }
     
+    @Override
+    public byte[] sign(byte[] contentBytes) throws SignatureGenerationException {
```
lbalmaceda: shouldn't this method also be annotated as `@deprecated` ?

lib/src/main/java/com/auth0/jwt/algorithms/NoneAlgorithm.java
```diff
@@ -24,4 +24,8 @@ public void verify(DecodedJWT jwt) throws SignatureVerificationException {
         return new byte[0];
     }
 
+    @Override
+    public byte[] sign(byte[] contentBytes) throws SignatureGenerationException {
```
lbalmaceda: shouldn't this method also be annotated as `@deprecated` ?

lib/src/main/java/com/auth0/jwt/algorithms/HMACAlgorithm.java
```diff
@@ -63,4 +63,12 @@ public void verify(DecodedJWT jwt) throws SignatureVerificationException {
         }
     }
 
+    @Override
+    public byte[] sign(byte[] contentBytes) throws SignatureGenerationException {
```
lbalmaceda: shouldn't this method also be annotated as `@deprecated` ?

skjolber: @Deprecated added

lib/src/main/java/com/auth0/jwt/algorithms/NoneAlgorithm.java
skjolber: @Deprecated added

lib/src/main/java/com/auth0/jwt/algorithms/RSAAlgorithm.java
skjolber: @Deprecated added

skjolber:
@lbalmaceda  Added a few more unit tests, Javadoc in CryptoHelper, @Deprecated to the old sign(..) methods, some cleanups.

lbalmaceda:
Thanks a lot for sticking with this 👍 👍

skjolber:
@lbalmaceda Thank you for your patience :+1: :1st_place_medal:

lbalmaceda:
@skjolber I forgot to ask you that the other day. Thanks! 💥

Issue #256, opened 2018-06-04
make validation more customizable


wuan:
We currently have a use case to validate that at least one audience of the token is matched by the list of possible audiences. As `JWTVerifier` is final this becomes a very ugly hack, because a lot of code has to be copied for a simple change.

What is your preferred way of making that possible:
  
  * making `JWTVerifier` non final
  * updating `JWTVerifier` to allow select verfications to be replaced, e. g. `assertValidAudienceClaim()`

lbalmaceda:
I'm considering making a lot of final classes non-final so tests and extensibility is possible. As of today, the behavior of the `withAudience` verification is the opposite: All of the given audiences must be present on the token as you might already realized. This has always been our intention since as we see it, you always know who the token is intended for. Using `withArrayClaim` has the same behavior. Changing this would be a breaking change as the default behavior is changing. However, there might be a way to introduce a boolean to this check so that the way it goes is inverted. e.g. "at least one audience from this list is on the token's audience claim". Making this latter change is not a priority for us, but I'll prioritize the first one and add a task in our backlog to review which classes could be made non-final in order to improve extensibility. I'll leave this open as a reminder. Thanks

rocketraman:
How about using this custom validator I just created? https://github.com/auth0/java-jwt/pull/290

PR #258, opened 2018-06-05
add verification feature mechanism and add at least one audience validation

wuan:
This PR adds a generic expandable feature set added to the JWTValidator. As an example a different audience validation `AT_LEAST_ONE_AUDIENCE_REQUIRED` was added.

This could be a possible solution to solve #256 in a generic way.


lbalmaceda:
This has been open for a long time now. We've evaluated it with the team and decided to not move forward with this request, as it's not something we plan to support in the short/mid-term. However, the use case you suggested around one audience match is a valid one. We are thinking to allow the extensibility of the validation logic.

PR #266, opened 2018-07-12
nbf is checked if it is in the past

PerfectPixel:
Just a small fix to the docs to avoid confusion. Unlike `exp` which must be in the future, `nbf` must be in the past.


lbalmaceda:
Superseded

PR #278, opened 2018-08-20
Support verification of Long[] datatype like in JWTCreator

skjolber:
Add Long array type so that the Verification claim types are the same as supported in JWTCreator.

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/interfaces/Verification.java
+++ b/lib/src/main/java/com/auth0/jwt/interfaces/Verification.java
@@ -161,14 +161,17 @@ public interface Verification {
     Verification withArrayClaim(String name, Integer... items) throws IllegalArgumentException;
 
     /**
-     * Skip the Issued At ("iat") date verification. By default, the verification is performed.
-     */
-    Verification ignoreIssuedAt();
-
-    /**
      * Creates a new and reusable instance of the JWTVerifier with the configuration already provided.
      *
      * @return a new JWTVerifier instance.
      */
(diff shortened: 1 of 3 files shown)


skjolber:
Ping.

lbalmaceda:
@skjolber please rebase+fix conflicts so I can leave a review

lib/src/main/java/com/auth0/jwt/interfaces/Verification.java
```diff
@@ -160,15 +160,18 @@
...
-     * Skip the Issued At ("iat") date verification. By default, the verification is performed.
-     */
-    Verification ignoreIssuedAt();
-
     /**
      * Creates a new and reusable instance of the JWTVerifier with the configuration already provided.
      *
      * @return a new JWTVerifier instance.
      */
+    Verification withArrayClaim(String name, Long ... items) throws IllegalArgumentException;
```
lbalmaceda: This change swapped the javadoc lines. 

- Make sure the new `withArrayClaim()` method has something like this:

```
/**
     * Require a specific Array Claim to contain at least the given items.
     *
     * @param name  the Claim's name.
     * @param items the items the Claim must contain.
     * @throws IllegalArgumentException if the name is null.
     */
```

- Make sure the `build()` method keeps this:
```
/**
     * Creates a new and reusable instance of the JWTVerifier with the configuration already provided.
     *
     * @return a new JWTVerifier instance.
     */
```

skjolber: Fixed

lbalmaceda:
Thanks for this @skjolber

Issue #279, opened 2018-08-21
Expose new new verify method signature.


KowalczykBartek:
Hi Auth0 team :) 
recently I found that it would be quite useful to have following method signature,
```java
public DecodedJWT verify(DecodedJWT decodedJWT)
```
I have use case for first, checking some stuff inside JWT token, then, in some cases I want to check signature - to not waste CPU cycles, it would be good to have verify method that accepts DecodedJWT instance.

Pull request with change is here : https://github.com/auth0/java-jwt/pull/280

Thanks !

martinoconnor:
This is useful when using JWKS but not `KeyProvider`. Unfortunately `KeyProvider` provides no means for dealing with exceptions that may occur when retrieving the key.

rocketraman:
How about using this custom validator I just created? https://github.com/auth0/java-jwt/pull/290

KowalczykBartek:
From my perspective, it would be simpler to just call method that already is there :)

martinoconnor:
@rocketraman a new custom validation would not solve the case where we need to read the contents of the JWT before we can verify it. This issue makes sense because we would not have to parse the JWT a second time.

martinoconnor:
@rocketraman no, sometimes you have to decode the JWT before you can verify it. If it is signed with the RS246 algorithm and the public key is available via JWKS, it is necessary to decode the JWT to retrieve the key ID so the public key can be looked up in the providers' well-known location for publishing JWKS public keys.

rocketraman:
@martinoconnor  Yes, but that is what the `KeyProvider` is for. I just used this myself to validate tokens from Microsoft Azure AD. That is acceptable because its a limited purpose hook, not a general "give me everything and let me decide what to do".

KowalczykBartek:
guys - @martinoconnor , @rocketraman - are you sure that someone still maintains that project ? XD

martinoconnor:
@rocketraman 
1. This PR is not for allowing the JWT to be decoded without verification, that functionality has long existed. This PR allows a previously decoded JWT to subsequently be verified. You should probably actually look at the PR source
2. The `KeyProvider` does not currently handle exceptions very well, for example, if there are network issues while accessing the JWKS, there is not really a good way to handle it yet

@KowalczykBartek I think @lbalmaceda maintains this project

rocketraman:
[...]

PR #280, opened 2018-08-21
Expose new verify method signature #279

KowalczykBartek:


lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -356,6 +356,23 @@ public DecodedJWT verify(String token) throws JWTVerificationException {
         return jwt;
```
martinoconnor: DRY: The existing method should delegate to the new method with the decoded token.

KowalczykBartek: done

lbalmaceda:
I'm still not convinced this should be merged. I'll leave it open and keep the discussion on the linked issue.

Issue #246, opened 2018-03-09
Validate against multiple issuers


cescoferraro:
Is there a way to validate against multiple issuers with the current api?

lbalmaceda:
No, since the `JWTVerifier#withIssuer`  only accepts a single string and the `JWTVerifier#withArrayClaim` will check that **all** the passed array contents are present in the token rather than only checking if the token's issuer is part of the passed array.
Could be added though.

crizo23:
Other JWT libraries accept an array of issuers so this would be a very nice-to-have.

jogee:
We are looking for something similar when switching to a custom domain with auth0, when you switch to a custom domain the issuer changes although the auth0 tenant stays the same. To handle this we temporarily have to accept multiple issuers.

The patch with having multiple verifiers seems to be the best solution to be able to transition properly.

lbalmaceda:
This is already in our backlog. I'll discuss with the team on the coming days if we should raise the priority. 👍

rocketraman:
I'd like to have this too. My use case is that we have one issuer prefix e.g. `https://foo.com` with multiple valid paths, such as `https://foo.com/v1-issuer` and `https://foo.com/v2-issuer`. Therefore, I need to be able to pass in a list of valid issuers.

You could make this even more generic by supporting a method like `withIssuer(IssuerValidator)` where IssuerValidator is a SAM interface for easy lamda implementation.

rocketraman:
> That being said, more complex cases like that can be handled at the application level after parsing, so I would be fine with just the multiple issuer support.

That being said, enabling this at the application level would be nice by supporting a generic validation method like this on `Verification`, where each call contributes an additional custom validation:

`withCustomValidation(CustomJwtValidator)` where `CustomJwtValidator` is:

```
interface `CustomJwtValidator` {
  validate(DecodedJWT jwt) throws JWTVerificationException;
}
```

rocketraman:
FYI, the logic I describe above implemented in https://github.com/auth0/java-jwt/pull/290.

jogee:
With the custom validator the issuer claim would still fail if multiple issuers aren't supported, these look like two separate concepts. Adding the possibility to add custom validations on application layer seems like a good idea though.

rocketraman:
> With the custom validator the issuer claim would still fail if multiple issuers aren't supported, these look like two separate concepts. Adding the possibility to add custom validations on application layer seems like a good idea though.

[...]

PR #288, opened 2018-10-05
Support multiple issuers #246

itdevelopmentapps:
Added support for multiple issuers.

I wanted to change the previous pull request so I closed it, maybe if you guys can remove that one in favor for this one. 

Many thanks.


lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -452,7 +453,13 @@ private void assertDateIsPast(Date date, long leeway, Date today) {
 
     private void assertValidAudienceClaim(List<String> audience, List<String> value) {
         if (audience == null || !audience.containsAll(value)) {
-            throw new InvalidClaimException("The Claim 'aud' value doesn't contain the required audience.");
+            throw new InvalidClaimException(String.format("The Claim '%s' value doesn't contain the required audience.", PublicClaims.AUDIENCE));
```
lbalmaceda: Given that this method is already dedicated to the audience claim, this change is redundant. Please rollback

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -452,7 +453,13 @@ private void assertDateIsPast(Date date, long leeway, Date today) {
...
+        }
+    }
+
+    private void assertValidIssuerClaim(String issuer, List<String> value) {
+        if (issuer == null || !value.contains(issuer)) {
+            throw new InvalidClaimException(String.format("The Claim '%s' value doesn't match the required issuer.", PublicClaims.ISSUER));
```
lbalmaceda: Given that this method is already dedicated to the issuer claim, calling String.format is redundant. Please rollback

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
```diff
@@ -557,12 +568,25 @@ public void shouldThrowOnInvalidJWTId() throws Exception {
     public void shouldRemoveClaimWhenPassingNull() throws Exception {
         Algorithm algorithm = mock(Algorithm.class);
         JWTVerifier verifier = JWTVerifier.init(algorithm)
-                .withIssuer("iss")
-                .withIssuer(null)
+                .withSubject("1234567890")
```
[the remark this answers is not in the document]
itdevelopmentapps: @lbalmaceda it's not possible to test this with `null` on issuer since I can't pass `null` to Arrays.asList() varargs parameter because it's `@NotNull` resulting in a NullPointerException.

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
```diff
@@ -557,12 +568,25 @@ public void shouldThrowOnInvalidJWTId() throws Exception {
     public void shouldRemoveClaimWhenPassingNull() throws Exception {
         Algorithm algorithm = mock(Algorithm.class);
         JWTVerifier verifier = JWTVerifier.init(algorithm)
-                .withIssuer("iss")
-                .withIssuer(null)
```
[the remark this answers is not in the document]
itdevelopmentapps: Resolving as duplicate.

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -58,12 +58,12 @@ static Verification init(Algorithm algorithm) throws IllegalArgumentException {
...
          *
-         * @param issuer the required Issuer value
+         * @param issuer the required Issuer value. If multiple values are given, the claim must at least match one of them
          * @return this same Verification instance.
          */
         @Override
-        public Verification withIssuer(String issuer) {
-            requireClaim(PublicClaims.ISSUER, issuer);
+        public Verification withIssuer(String... issuer) {
+            requireClaim(PublicClaims.ISSUER, Arrays.asList(issuer));
```
lbalmaceda: You should not need to call `Arrays.asList` at all. If the varargs were used instead of a String, it would be an instance of Array which can already be considered a "list".

Similarly, if you could go ahead and fix the `withAudience` method which shouldn't be calling `Arrays.asList` as well. (same reason). Feel free to do it here or in a separate PR if desired.

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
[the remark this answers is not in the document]
lbalmaceda: Once you fix the comment I left above about the `Arrays.asList` you can rollback this to how it was before.

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -90,7 +90,7 @@ public Verification withSubject(String subject) {
          */
         @Override
         public Verification withAudience(String... audience) {
-            requireClaim(PublicClaims.AUDIENCE, Arrays.asList(audience));
+            requireClaim(PublicClaims.AUDIENCE, audience);
```
lbalmaceda: Let's use this instead

```suggestion
            requireClaim(PublicClaims.AUDIENCE, audience == null ? null : Arrays.asList(audience));
```

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -61,11 +61,11 @@ static Verification init(Algorithm algorithm) throws IllegalArgumentException {
...
          * Require a specific Issuer ("iss") claim.
          *
-         * @param issuer the required Issuer value
+         * @param issuer the required Issuer value. If multiple values are given, the claim must at least match one of them
          * @return this same Verification instance.
          */
         @Override
-        public Verification withIssuer(String issuer) {
+        public Verification withIssuer(String... issuer) {
             requireClaim(PublicClaims.ISSUER, issuer);
```
lbalmaceda: Let's use this instead 

```suggestion
             requireClaim(PublicClaims.ISSUER, issuer == null ? null : Arrays.asList(issuer));
```

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -398,8 +398,7 @@ private void verifyClaims(DecodedJWT jwt, Map<String, Object> claims) throws Tok
         for (Map.Entry<String, Object> entry : claims.entrySet()) {
             switch (entry.getKey()) {
                 case PublicClaims.AUDIENCE:
-                    //noinspection unchecked
-                    assertValidAudienceClaim(jwt.getAudience(), (List<String>) entry.getValue());
+                    assertValidAudienceClaim(jwt.getAudience(), Arrays.asList((String[]) entry.getValue()));
```
lbalmaceda: Rolling back this
```suggestion
                    assertValidAudienceClaim(jwt.getAudience(), (List<String>) entry.getValue());
```

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -411,7 +410,7 @@ private void verifyClaims(DecodedJWT jwt, Map<String, Object> claims) throws Tok
                     assertValidDateClaim(jwt.getNotBefore(), (Long) entry.getValue(), false);
                     break;
                 case PublicClaims.ISSUER:
-                    assertValidStringClaim(entry.getKey(), jwt.getIssuer(), (String) entry.getValue());
+                    assertValidIssuerClaim(jwt.getIssuer(), Arrays.asList((String[]) entry.getValue()));
```
lbalmaceda: Rolling back this

```suggestion
                    assertValidIssuerClaim(jwt.getIssuer(), (List<String>) entry.getValue());
```

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
```diff
@@ -587,7 +599,20 @@ public void shouldRemoveClaimWhenPassingNull() throws Exception {
                 .build();
 
         assertThat(verifier.claims, is(notNullValue()));
-        assertThat(verifier.claims, not(hasKey("iss")));
+        assertThat(verifier.claims, not(hasKey("sub")));
```
lbalmaceda: This should remain the same. When we pass `null` we're removing that claim check.

```suggestion
        assertThat(verifier.claims, not(hasKey("iss")));
```

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
```diff
@@ -587,7 +599,20 @@ public void shouldRemoveClaimWhenPassingNull() throws Exception {
...
+    }
+
+    @Test
+    public void shouldThrowWhenNoIssuerPresent() {
```
lbalmaceda: Remove this test as it's already present here `shouldThrowOnInvalidIssuer()`

lbalmaceda:
@itdevelopmentapps are you still interesting in merging this PR? Please, don't hesitate to ask any questions.

itdevelopmentapps:
@lbalmaceda please advise in what test you would like to see to increase coverage so that we can get this pull request accepted.

lbalmaceda:
By checking the [coverage report](https://codecov.io/gh/auth0/java-jwt/compare/7fcd1bfece8197f0c44cd020a03045df6cf182de...c77b7e7b48022fa011608529410e613970a9ab70/diff), there are 2 missing branches.
- `withAudience` needs to be called both with a valid audience and a null string value. Then you need to call `verify`. Tests are only present for a "valid audience value" [here](https://github.com/auth0/java-jwt/blob/c77b7e7b48022fa011608529410e613970a9ab70/lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java#L99-L139). 
- for `withIssuer` tests are present for "both valid issuer and null string values" [here](https://github.com/auth0/java-jwt/blob/c77b7e7b48022fa011608529410e613970a9ab70/lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java#L43-L75) but you need to add another one that verifies against a JWT payload that doesn't have an issuer claim defined.

I hope that's clear enough. It should help make the check green.

itdevelopmentapps:
@lbalmaceda we've added the suggested tests.

lbalmaceda:
Thank you. Will release before eow.

itdevelopmentapps:
Many thanks.

vikeri:
Using Clojure this was a breaking change, got the following error:
```
java.lang.ClassCastException: java.lang.String cannot be cast to [Ljava.lang.String;
```
When trying to call `withIssuer`.
To fix:
Before
```clojure
(.withIssuer Verification "example")
```
After
```clojure
(.withIssuer Verification (into-array String ["example]))
```

PR #289, opened 2018-10-10
Support to add objects (primitive and collection) as claim in token pa…

complanboy2:
…yload

As requested @gustavoamigo we are giving support to add objects if primitive and collection types as claim in token payload.

Followed the suggestion by @lbalmaceda to make this change.
https://github.com/auth0/java-jwt/pull/164#issuecomment-295764851

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Change Log
 
+## [3.5.0](https://github.com/auth0/java-jwt/tree/3.5.0) (2019-01-03)
+[Full Changelog](https://github.com/auth0/java-jwt/compare/3.4.1...3.5.0)
+
+**Added**
+- Verify a DecodedJWT  [\#308](https://github.com/auth0/java-jwt/pull/308) ([martinoconnor](https://github.com/martinoconnor))
+
+**Changed**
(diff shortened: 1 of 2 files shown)


complanboy2:
@lbalmaceda waiting for your review. Please review at your free time.

complanboy2:
> The added `withClaim()` method is not type safe, and it is not possible to overload it due to type erasure. Also, nested collections would be raw types, so there would be no guarantee as to the types they would contain.
> 
> A reasonable compromise may be to accept only primitive wrapper types, `Set`s and `List`s and to call `toString()` on anything else. That would require `instanceof` checks at runtime which may be ugly.

Hi @martinoconnor , thank you for the reivew. It helped fixing it better.
Please review the updated code when you have time.

ornom0n:
Hello,

Adding this feature would allow me to migrate to JWT 3 as one of the claims we use is a Map and that cannot be changed. Is anything pending here?

complanboy2:
> Hello,
> 
> Adding this feature would allow me to migrate from JWT 2 to JWT 3 as one of the claims we use is a Map and that cannot be changed. Is anything pending here?

Hi OrnomOn,

I'm waiting on review, once the review is done, we are good to merge.
If it is helpful to you, please go ahead with the use. And please have a look at checks once, though major are passed.

complanboy2:
Any hope for getting this reviewed and also please check, #297 as well.

Issue #208, opened 2017-10-19
[PATCH] JWTVerifier: allow verify parsed tokens


jcodagnone:
Today you can't parse the token once and verify it against N validators
(based on the algorithm or another feature…).

Decoding the token is not cheap in general terms (there are lots of
strings instantiations)

[0001-JWTVerifier-allow-verify-parsed-tokens.txt](https://github.com/auth0/java-jwt/files/1400032/0001-JWTVerifier-allow-verify-parsed-tokens.txt)

kaandok:
@lbalmaceda hi, any plans to merge this in/any help required?

filip26:
note this patch is workaround for #246 because allows to have multiple different verifiers.

e.g.
1. decode token
2. use decoded token property (e.g. issuer) to select appropriate verifier
3. call verifier.verify(decoded token)

rocketraman:
I created a custom validator that can be used effectively for this and other use cases: https://github.com/auth0/java-jwt/pull/290

rocketraman:
@AndreasHuber-CH Replying to your comment here:

> Hi. This looks nice, but it doesn't cover #208 entirely. For example it is still not possible to chose the algorithm based on the JWT header.

Yes, the algorithm is the special case because passing it is right now seems to be the only way to obtain a `Verification`. Note that I believe it was done this way to avoid a security risk -- if you chose the algorithm based on the token itself, then you are opening yourself up to a "none" algorithm: see https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/.

lbalmaceda:
Was added here https://github.com/auth0/java-jwt/issues/279

PR #290, opened 2018-10-14
Support custom validation

rocketraman:
Adds the ability for users to add one or more custom validations to the
JWT verification process, by using `withCustomValidation` on the
validation builder.

This is useful to workaround the inability to perform particular
validations directly with the library, such as issue #19, or to do other
validations that are not possible with the built-in validations.


rocketraman:
I believe this should help close 5 issues, with a pretty simple change.

AndreasHuber-CH:
Hi. This looks nice, but it doesn't cover #208 entirely. For example it is still not possible to chose the algorithm based on the JWT header.

rocketraman:
Re-open please. Not sure why this was not reviewed.

ssgao:
@auth0/dx-sdks-approver would it be possible to re-open this one? It's a very helpful PR.

PR #296, opened 2018-10-29
Remove unnecessary cast between long/double and floor call

jhorstmann:
Calculating `time / 1000 * 1000` already returns the correctly truncated result using long arithmetic


lbalmaceda:
I think this was originally added either because the JDK7 compiler complained or because tests were randomly failing because of that +/- second change in the verification.

jhorstmann:
I could not find the historical commit that introduced the `floor` call, but if the tests were flaky that must have been for a different reason. The division and multiplication are happing with `long` type, so there is no fractional part that would be removed by the floor call.

Issue #254, opened 2018-05-29
`iat` should not be used for token lifetime validation


nicosabena:
The `exp` and `nbf` claims should be used to validate the token lifetime, but the `iat` should be treated as merely informative.
The [JWT spec](https://tools.ietf.org/html/rfc7519#section-4.1.5) does not indicate any validation for `iat`. Other libraries like .Net also don't use `iat` (e.g. [here](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/a07d6f949dca3d26d024744d8d80c2bcfb6dd2f3/src/System.IdentityModel.Tokens.Jwt/JwtSecurityTokenHandler.cs#L785)).

wuan:
We had issues with clock skew between servers which resulted in invalid tokens in rare cases. The logic added to our verifier which now sets a default leeway if no leeway is set could be removed if `iat` is not verified to be in the past.

DinoChiesa:
Irrespective of clock skew (which can be handled by leeway), the spec has nothing to say about whether iat ought to be considered during validation.  One may reasonably take either view: it should be considered, or it should be ignored. 

On the side of "should be ignored" - already we have  nbf and exp. These are validity times and are described as such.  Therefore using iat as well as these properties would be redundant. By this reasoning the most sensible option is to ignore iat. 

On the side of "should be considered" - common sense tells us that iat cannot be in the future. Any token with iat in the future looks bogus. 

Libraries are free to adopt either view.  java-jwt has taken the view that iat should be considered during validation. But users of libraries might differ, and it doesn't make sense to lock people into a view, given the silence in the spec on this point.  

Therefore, a simple solution: Introduce a new method on the JWTVerifier class, ignoreIssuedAt().  This calls requireClaim(ISSUED_AT, null). Then also modify the verifyClaims() method and check for null on the ISSUED_AT claim, in which case no-op.

gedl:
Regardless of the final decision, if the `iat` check is to assert if `now` is not before `iat`, then the implementation is fundamentally wrong as it stands because it also rejects as invalid dates that are exactly `now`:

```
  case PublicClaims.ISSUED_AT:
                    assertValidDateClaim(jwt.getIssuedAt(), (Long) entry.getValue(), false);
```

which calls
```
  private void assertDateIsPast(Date date, long leeway, Date today) {
        today.setTime(today.getTime() + leeway * 1000);
        if (date != null && today.before(date)) {
[...]

PR #297, opened 2018-10-30
Allow to skip "issued at" validation

complanboy2:
The default behavior remains the same: Always verify the iat.

Will close https://github.com/auth0/java-jwt/issues/254


complanboy2:
Any hop for getting this reviewed in 2018 !

And also please check, #289

svenwb:
Is there any eta when this will be reviewed?

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
```diff
@@ -494,6 +494,33 @@ public void shouldValidateIssuedAtWithLeeway() throws Exception {
         assertThat(jwt, is(notNullValue()));
     }
 
+	// Issued At with future date
+	@Test (expected = InvalidClaimException.class)
```
lbalmaceda: This test is already present under the name `shouldThrowOnInvalidIssuedAtIfPresent`. Please remove it.

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
```diff
@@ -494,6 +494,33 @@ public void shouldValidateIssuedAtWithLeeway() throws Exception {
...
+		String token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE0Nzc1OTJ9.CWq-6pUXl1bFg81vqOUZbZrheO2kUBd2Xr3FUZmvudE";
+		JWTVerifier.BaseVerification verification = (JWTVerifier.BaseVerification) JWTVerifier.init(Algorithm.HMAC256("secret"));
+
+		DecodedJWT jwt = verification.build(clock).verify(token);
+		assertThat(jwt, is(notNullValue()));
+	}
+
+	// Issued At with future date and ignore flag
+	@Test
+	public void shouldNotVerifyIATOnIgnoreIssuedAt() throws Exception {
```
lbalmaceda: Rename to `shouldSkipIssuedAtVerificationWhenFlagIsPassed`

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
```diff
@@ -494,6 +494,33 @@ public void shouldValidateIssuedAtWithLeeway() throws Exception {
...
+		when(clock.getToday()).thenReturn(new Date(DATE_TOKEN_MS_VALUE - 1000));
+
+		String token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpYXQiOjE0Nzc1OTJ9.CWq-6pUXl1bFg81vqOUZbZrheO2kUBd2Xr3FUZmvudE";
+		JWTVerifier.BaseVerification verification = (JWTVerifier.BaseVerification) JWTVerifier.init(Algorithm.HMAC256("secret"));
+		verification.ignoreIssuedAt();
+
+		DecodedJWT jwt = verification.build(clock).verify(token);
+		assertThat(jwt, is(notNullValue()));
+	}
+
```
lbalmaceda: Please add another test that calls `acceptsIssuedAt(..)` passing a value and then `ignoreIssuedAt()`. This test would expect no exceptions. You'll see how this test would fail unless you put the claim check in the right place (see my review's general comment)

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -150,6 +151,10 @@ public Verification acceptIssuedAt(long leeway) throws IllegalArgumentException
             return this;
         }
 
+		public void ignoreIssuedAt() {
```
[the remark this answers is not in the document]
complanboy2: done, thank you.

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
complanboy2: done, thank you.

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
complanboy2: done, thank you.

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
complanboy2: Understand and agree with you on acceptsIssuedAt()

And since the flag is available at builder level, try to achieve the job at the claims map level rather at verification level later in parent class. Please see the updated code. Thank you.

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -316,17 +325,20 @@ private void assertNonNull(String name) {
...
+			if (!claims.containsKey(PublicClaims.EXPIRES_AT)) {
+				claims.put(PublicClaims.EXPIRES_AT, defaultLeeway);
+			}
+			if (!claims.containsKey(PublicClaims.NOT_BEFORE)) {
+				claims.put(PublicClaims.NOT_BEFORE, defaultLeeway);
+			}
+			if (!claims.containsKey(PublicClaims.ISSUED_AT)) {
+				claims.put(PublicClaims.ISSUED_AT, defaultLeeway);
+			}
+			if(ignoreIssuedAt) {
```
lbalmaceda: swap this if with the one above so you avoid the case in which the claim is added and immediately removed. You could also `return` right after removing the claim, so the following if is not considered.

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -150,6 +151,14 @@ public Verification acceptIssuedAt(long leeway) throws IllegalArgumentException
             return this;
         }
 
+        /**
+         * Call this method to skip the default issued_at verification
```
lbalmaceda: Please change to

```suggestion
         * Skip the Issued At ("iat") date verification. By default, the verification is performed.
         
         * @return this same Verification instance.

```

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -316,17 +325,20 @@ private void assertNonNull(String name) {
...
-                claims.put(PublicClaims.EXPIRES_AT, defaultLeeway);
-            }
-            if (!claims.containsKey(PublicClaims.NOT_BEFORE)) {
-                claims.put(PublicClaims.NOT_BEFORE, defaultLeeway);
-            }
-            if (!claims.containsKey(PublicClaims.ISSUED_AT)) {
-                claims.put(PublicClaims.ISSUED_AT, defaultLeeway);
-            }
-        }
+		private void addLeewayToDateClaims() {
```
lbalmaceda: this method is still not formatted according to the rest of the file. Please run the code formatter or do it manually. Thanks

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
complanboy2: done, thank you.

lbalmaceda:
Thanks!

lbalmaceda:
@svenwb I'll see if I can fit another change by end of week and make a release.

PR #320, opened 2019-01-24
Fixing JwtCreator builder when setting headers as a map

maxbalan:
currently when using JwtCreator when setting headers as a map it is overwriting existing headers map, which forces to use `withHeader(Map<String, Object> headerClaims)` first then any other headers.

e.g.
1. the key id is missing if I do `JWT.create().withKeyId(ID).withHeader(map)`
2. the key id is not missing if I do `JWT.create().withHeader(map).withKeyId(ID)`

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -67,14 +67,15 @@ public final class JWTCreator {
         /**
          * Add specific Claims to set as the Header.
          * If provided map is null then nothing is changed
-         * If provided map contains a header with null value then that header will be removed from the header claims
+         * If provided map contains a claim with null value then that claim will be removed from the header
          *
          * @param headerClaims the values to use as Claims in the token's Header.
          * @return this same Builder instance.
          */
         public Builder withHeader(Map<String, Object> headerClaims) {
-            if (headerClaims == null)
+            if (headerClaims == null) {
                 return this;
+            }
 
             for (Map.Entry<String, Object> entry : headerClaims.entrySet()) {
                 if (entry.getValue() == null) {
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -66,12 +66,24 @@ public final class JWTCreator {
 
         /**
          * Add specific Claims to set as the Header.
+         * If provided map is null then nothing
+         * If provided map contains a header with null value then that header will be removed from the header claims
          *
          * @param headerClaims the values to use as Claims in the token's Header.
          * @return this same Builder instance.
          */
         public Builder withHeader(Map<String, Object> headerClaims) {
-            this.headerClaims.putAll(headerClaims);
+            if (headerClaims == null)
+                return this;
+
+            for (Map.Entry<String, Object> entry : headerClaims.entrySet()) {
+                if (entry.getValue() == null) {
+                    this.headerClaims.remove(entry.getKey());
+                } else {
+                    this.headerClaims.put(entry.getKey(), entry.getValue());
+                }
+            }
+
             return this;
         }
 
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -66,7 +66,7 @@ public final class JWTCreator {
 
         /**
          * Add specific Claims to set as the Header.
-         * If provided map is null then nothing
+         * If provided map is null then nothing is changed
          * If provided map contains a header with null value then that header will be removed from the header claims
          *
          * @param headerClaims the values to use as Claims in the token's Header.
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -71,7 +71,7 @@ public final class JWTCreator {
          * @return this same Builder instance.
          */
         public Builder withHeader(Map<String, Object> headerClaims) {
-            this.headerClaims = new HashMap<>(headerClaims);
+            this.headerClaims.putAll(headerClaims);
             return this;
         }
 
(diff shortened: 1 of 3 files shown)


maxbalan:
@lbalmaceda thank you for a quick reply, this makes sense, I will update it and commit a new change

maxbalan:
@lbalmaceda could you please have another look at this change? I will really appreciate if we can have this fixed in the next version. thanks.

lbalmaceda:
@maxbalan 😐Sorry I missed this. I've set up a reminder and will review it tomorrow.

lib/src/main/java/com/auth0/jwt/JWTCreator.java
```diff
@@ -66,12 +66,24 @@ private JWTCreator(Algorithm algorithm, Map<String, Object> headerClaims, Map<St
 
         /**
          * Add specific Claims to set as the Header.
+         * If provided map is null then nothing is changed
+         * If provided map contains a header with null value then that header will be removed from the header claims
```
lbalmaceda: Small rewrite
```suggestion
         * If provided map contains a claim with null value then that claim will be removed from the header
```

lib/src/main/java/com/auth0/jwt/JWTCreator.java
```diff
@@ -66,12 +66,24 @@ private JWTCreator(Algorithm algorithm, Map<String, Object> headerClaims, Map<St
...
          *
          * @param headerClaims the values to use as Claims in the token's Header.
          * @return this same Builder instance.
          */
         public Builder withHeader(Map<String, Object> headerClaims) {
-            this.headerClaims = new HashMap<>(headerClaims);
+            if (headerClaims == null)
```
lbalmaceda: Please use braces `{}` for this even if it's a one liner, to follow the project code style

lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
```diff
@@ -53,6 +53,49 @@ public void shouldAddHeaderClaim() throws Exception {
...
+
+        String signed = JWTCreator.init()
+                                  .withClaim("test", 123)
+                                  .withHeader(header)
+                                  .sign(Algorithm.HMAC256("secret"));
+
+        assertThat(signed, is(notNullValue()));
+        String[] parts = signed.split("\\.");
+        String headerJson = new String(Base64.decodeBase64(parts[0]), StandardCharsets.UTF_8);
+        assertThat(headerJson, JsonMatcher.hasEntry("test", null));
```
lbalmaceda: The final JSON object shouldn't have the key at all. Is not correct for it to have the claim present with a null value. e.g. if you convert this back to a Map, the key shouldn't be part of that Map.

lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
```diff
@@ -53,6 +53,49 @@ public void shouldAddHeaderClaim() throws Exception {
...
+    public void shouldReturnBuilderIfNullMapIsProvided() throws Exception {
+        String signed = JWTCreator.init()
+                                  .withHeader(null)
+                                  .sign(Algorithm.HMAC256("secret"));
+
+        assertThat(signed, is(notNullValue()));
+    }
+
+    @Test
+    public void shouldOverwriteExistingIfHeadersMapContainsTheSameKey() throws Exception {
```
lbalmaceda: This test looks confusing as the name tries to tell that a header claim with the same name was already set and then overwritten by the `withHeader` call. However that is not true, as the `withClaim` is adding a claim to the _payload_ object (not the header). There's no overwrite here. 

You might want to review the PR description where you showed the failing scenario with the KID claim and test that instead.


```java
Map<String, Object> header = new HashMap<String, Object>();
header.put("kid", "123456");


String signed = JWTCreator.init()
                                  .withKeyId("abcdef")
                                  .withHeader(header)

//Check header.kid = 123456
```

You can even duplicate this test and check that when called in the alternate order, header.kid is "abcdef". e.g. name the test something like "should respect order of builder calls when setting the header"

lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
```diff
@@ -53,6 +53,49 @@ public void shouldAddHeaderClaim() throws Exception {
...
+        assertThat(signed, is(notNullValue()));
+        String[] parts = signed.split("\\.");
+        String headerJson = new String(Base64.decodeBase64(parts[0]), StandardCharsets.UTF_8);
+        assertThat(headerJson, JsonMatcher.hasEntry("test", 456));
+    }
+
+    @Test
+    public void shouldRemoveHeaderIfTheValueIsNull() throws Exception {
+        Map<String, Object> header = new HashMap<String, Object>();
+        header.put("test", null);
```
lbalmaceda: The claim must first be set in order to be unset. Use a claim that you know is being set to the header.

maxbalan:
@lbalmaceda Hi, I fixed the pr as per your suggestions I hope this removes some of the ambiguity and adds more to tests readability, please have another look

lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
```diff
@@ -53,6 +54,65 @@ public void shouldAddHeaderClaim() throws Exception {
...
+    public void shouldReturnBuilderIfNullMapIsProvided() throws Exception {
+        String signed = JWTCreator.init()
+                                  .withHeader(null)
+                                  .sign(Algorithm.HMAC256("secret"));
+
+        assertThat(signed, is(notNullValue()));
+    }
+
+    @Test
+    public void shouldOverwriteExistingIfHeadersMapContainsTheSameKey() throws Exception {
```
lbalmaceda:
```suggestion
    public void shouldOverwriteExistingHeaderIfHeaderMapContainsTheSameKey() throws Exception {
```

maxbalan:
@lbalmaceda looks like when I applied your suggestions for test name change it dismissed the pr approve

PR #318, opened 2019-01-23
Support to add objects (primitive and collection) as claims in token …

complanboy2:
…payload

As requested @gustavoamigo we are giving support to add objects if primitive and collection types as claim in token payload.

Followed the suggestion by @lbalmaceda to make this change.
#164 (comment)

This PR is in continution to the old PR #289 which was accidentally closed. Please continue the review, thank you.

PR #322, opened 2019-02-01
Add map as claim

andy-shi88:
Currently when using the jwt builder, we need to use different single implementation of `withClaim` method, passing `[int, string, long...]`.

I found it hard to do abstraction if I have few payload to pass in. I created `withCustomClaim` that receive `Map` as argument. With this we can abstract the jwt generation with:

```
public String generateJWT(Map<String, Object> payload) {
        HashMap<String, Object> payload = new HashMap<>();
        payload.put("name", "andy");
        payload.put("age", 24);
        payload.put("isAdmin", true);

        JWTCreator.Builder builder = JWT.create().withCustomClaim(payload);
        return builder.sign(Algorithm.HMAC256("secret"));
}
```


lbalmaceda:
I appreciate the contribution. However there are many PRs already opened for the same feature and I can't spend enough time to review them all. I'd would like to accept one of the solutions that stick to my original requirements (discussed in a related issue).
By reading the description I think this is a candidate and will be my next PR review. https://github.com/auth0/java-jwt/pull/318.
I'll proceed to close this as duplicate. In the meantime, if this works for you, keep using the fork.

Issue #354, opened 2019-09-10
Provide a convenience method to verify a token without throwing Exception


smojum:
### Description
Currently, we use the following code to decrypt, but as we are supporting multiple data, if we cannot decrypt, we send the original value back.

```
        try {
            DecodedJWT decodedJWT = JWT.require(algorithm)
                .build()
                .verify(sid);
            return decodedJWT.getClaim(OLD_STYLE_SID).asString();
        } catch (Exception e) {
            LOGGER.info("Could not extract the SID, assuming old-style SID is being sent, returning the same.");
            return sid;
        }
```
This works fine, but we are seeing JWTDecodeException in Dynatrace, as Dynatrace instrumentation captures this exception and reports, which is creating noise. 

The ask if a convenience method (isValid) can be provided which will check the validity of the incoming string without throwing an exception.

I can work on this issue but need to ensure that it follows design guidelines.

### Prerequisites

- [x] I have checked the documentation for this library in the README.
- [x] I have checked the [Auth0 Community](https://community.auth0.com/) for related posts.
- [x] I have checked for related or duplicate [Issues](https://github.com/auth0/java-jwt/issues) and [PRs](https://github.com/auth0/java-jwt/pulls).
- [x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md).
- [x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).
- [x] I am reporting this to the correct repository (note this library is used by many libraries, such as [auth0-java](https://github.com/auth0/auth0-java)).

### Environment

Please provide the following:

- Version of this library used: 3.8.1
- Version of Java framework used: 1.8
- Additional libraries that might be affecting your instance: None

### Reproduction
See above for code sample

lbalmaceda:
@smojum I don't agree this is needed, as it's just a convenience method for your particular use case. 
You can add that as a helper/utility method in your application if you plan to ignore the exception, which has valuable information BTW.

In the case of the snippet you showed above, the reasons it could fail are:
- Invalid algorithm name
- Invalid signature
- Token `exp` indicates token has expired

In any failure case listed above, the claims should **not** be used, thus why it throws and fails. Ignoring the exception and returning the string token anyway does not seem right. 

**Can you expand on why you consume the invalid token anyway?**

[...]

PR #358, opened 2019-10-11
Add JWTVerifer added isValid method

smojum:
JWTVerifer added isValid method. Reason for the addition
1) it shows false exception for instrumented tools like Dynatrace
2) Functional programming without exception

### References

#354 https://github.com/auth0/java-jwt/issues/354

### Testing

Junit tests have been added. Test coverage has been verified.
- [X] This change adds test coverage
- [X] This change has been tested on the latest version of Java or why not

### Checklist

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
@@ -388,6 +388,26 @@ public final class JWTVerifier implements com.auth0.jwt.interfaces.JWTVerifier {
         return jwt;
     }
 
+    @Override
+    public Boolean isValid(String token) {
+        try {
+            verify(token);
+            return Boolean.TRUE;
+        } catch (Exception e) {
+            return Boolean.FALSE;
+        }
+    }
+
+    @Override
+    public Boolean isValid(DecodedJWT jwt) {
+        try {
+            verify(jwt);
+            return Boolean.TRUE;
+        } catch (Exception e) {
+            return Boolean.FALSE;
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -388,6 +388,26 @@ public DecodedJWT verify(DecodedJWT jwt) throws JWTVerificationException {
...
+        try {
+            verify(token);
+            return Boolean.TRUE;
+        } catch (Exception e) {
+            return Boolean.FALSE;
+        }
+    }
+
+    @Override
+    public Boolean isValid(DecodedJWT jwt) {
```
joshcanhelp: When would a decoded JWT be used? As in, why would someone pass in an already decoded token here? What would the `verify()` function check?

smojum: the reason I added the isValid(DecodedJWT jwt) just to keep it consistent with verify methods. As in verify method take either token or DecodedJWT.

smojum: I may have misunderstood your question. the reason behind to add this method is to 
1) the equivalent method verify throws JWTException which is not very functional programming friendly
2) when Exception is thrown some monitoring tools which does the instrumentation of code to see for any exceptions are showing a lot of JWT exception, when the intent in the code is just to verify whether it is a valid JWT token.

smojum:
Hi, 

Just checking, is there anything else needed from me.

lbalmaceda:
Not needed. Left a comment on the linked issue to discuss.

lbalmaceda:
Closing since it doesn't align with how we want this library to be consumed. The reason the verification fail should be relevant to the developer, not ignored.

PR #362, opened 2019-10-21
Automatic module name property in manifest

skjolber:
### Changes

Add a manifest entry for automatic module name as a small improvement for use with modules and JDK9+. Or better yet, use a plugin to create a release compatible with both 8 and 9+, for example a multi-release jar.

### References

http://branchandbound.net/blog/java/2017/12/automatic-module-name/

### Testing

Manually inspected resulting jar.

### Checklist

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -56,3 +56,13 @@ test {
         exceptionFormat "short"
     }
 }
+
+ext.moduleName = "com.auth0.jwt"
+
+jar {
+    inputs.property("moduleName", moduleName)
+
+    manifest {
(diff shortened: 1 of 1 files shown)


lbalmaceda:
We will look into what this means and if necessary, how we can include it on our release plugin.

skjolber:
Perhaps fix a multi-version jar instead? This is the quick-fix / minimum solution.

lbalmaceda:
Not a priority for the team today, but if done, will be done on the oss release library we use on every java library. I've created an issue there to track it and keep the conversation in a single place.

PR #403, opened 2020-03-06
Deprecate Date and array based claim methods

jimmyjames:
### Changes

Deprecate `java.util.Date` methods (in favor of `java.time.Instant`) and array-based claims (in favor of List-based claims).

### Checklist

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -147,8 +147,9 @@ public final class JWTCreator {
          *
          * @param expiresAt the Expires At value.
          * @return this same Builder instance.
+         * @deprecated Use {@linkplain #withExpiresAt(Instant)} instead.
          */
-        // TODO - Deprecate this method in favor of withExpiresAtInstant
+        @Deprecated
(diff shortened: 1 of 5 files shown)


lib/src/main/java/com/auth0/jwt/JWTCreator.java
```diff
@@ -312,7 +316,9 @@ public Builder withClaim(String name, Date value) throws IllegalArgumentExceptio
          * @param items the Claim's value.
          * @return this same Builder instance.
          * @throws IllegalArgumentException if the name is null.
+         * @deprecated Use {@linkplain #withClaim(String, List)} instead.
          */
+        @Deprecated
         public Builder withArrayClaim(String name, String[] items) throws IllegalArgumentException {
```
lbalmaceda: Do we provide an alternate way on **verification** for asserting lists claims? See the not deprecated method below
     
```
Verification withArrayClaim(String name, String... items) throws IllegalArgumentException;
```

jimmyjames: We do not. It's probably something we should re-evaluate though. And if we do add those (and later deprecate these), we should do that in a separate PR.

Issue #404, opened 2020-03-10
Issue on setting audience/issuer as var-args (JWTVerifier)


ssrm:
**Please do not report security vulnerabilities here**. The [Responsible Disclosure Program](https://auth0.com/whitehat) details the procedure for disclosing security issues.

**Thank you in advance for helping us to improve this library!** Please read through the template below and answer all relevant questions. Your additional work here is greatly appreciated and will help us respond as quickly as possible. For general support or usage questions, use the [Auth0 Community](https://community.auth0.com/) or [Auth0 Support](https://support.auth0.com/). Finally, to avoid duplicates, please search existing Issues before submitting one here.

By submitting an Issue to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

> when sending null to withAudience/withIssuer (JWTVerifier), varargs creates an array instead of null. Due to that even though there is no "aud/iss" value, it creates claim and it tries to match.

### Reproduction

https://stackoverflow.com/questions/4028059/calling-java-varargs-method-with-single-null-argument/4028113

### Environment

> Please provide the following:

- **Version of this library used:** 3.8.3
- **Version of the platform or framework used, if applicable:**
- **Other relevant versions (language, server software, OS, browser):**
- **Other modules/plugins/libraries that might be involved:**

PR #405, opened 2020-03-10
Added additional conditions on varargs to make sure given varargs or …

ssrm:
…not null

### Changes

If there is no aud/iss field in token, its a breaking change for the apps. In my case, we do not maintain audience field in token. Let me know if this is ok.

### References

- support ticket:  https://github.com/auth0/java-jwt/issues/404
- community post
- StackOverflow post
- support forum thread

### Testing

Please describe how this can be tested by reviewers. Be specific about anything not tested and reasons why. If this library has unit and/or integration testing, tests should be added for new functionality and existing tests should complete without errors. 

### Checklist

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
@@ -57,7 +57,7 @@ public final class JWTVerifier implements com.auth0.jwt.interfaces.JWTVerifier {
 
         @Override
         public Verification withIssuer(String... issuer) {
-            requireClaim(PublicClaims.ISSUER, issuer == null ? null : Arrays.asList(issuer));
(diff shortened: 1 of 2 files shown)


lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -230,6 +230,20 @@ private void requireClaim(String name, Object value) {
         }
     }
 
+    private static boolean checkVarargNull(String[] args) {
+        if (args == null || args.length == 0) {
+            return true;
+        }
+        boolean isAllNull = true;
+        for (String arg: args) {
+            if (arg != null && arg.trim().length() != 0) {
```
jimmyjames: Let's not check for empty strings here, just null. That will be consistent with how we handle empty strings for other claims.

jimmyjames:
I've taken the basis of this PR with another PR that is similar, except it does allow empty string values. #412

PR #406, opened 2020-03-11
Fix date serialization for list/map claims

jimmyjames:
### Changes

Custom list and map claims may include `java.util.Date` as well as `java.time.Instant` values for `NumericDate` values. When serializing these lists/maps, we need to account for `Date` values and serialize to the number of seconds since the epoch.

### References

[JWT Specification - NumericDate definition](https://tools.ietf.org/html/rfc7519#page-4)

### Testing

Relevant tests updated to expect `java.util.Date` claims to be serialized to the number of seconds since the epoch.

### Checklist

### DIFF
--- a/lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
+++ b/lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
@@ -2,6 +2,7 @@ package com.auth0.jwt;
 
 import com.auth0.jwt.algorithms.Algorithm;
 import com.auth0.jwt.impl.PublicClaims;
+import com.auth0.jwt.interfaces.DecodedJWT;
 import com.auth0.jwt.interfaces.ECDSAKeyProvider;
 import com.auth0.jwt.interfaces.RSAKeyProvider;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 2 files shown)


lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
```diff
@@ -530,10 +531,9 @@ public void shouldAcceptCustomMapClaimOfBasicObjectTypes() throws Exception {
 
         assertThat(jwt, is(notNullValue()));
         String[] parts = jwt.split("\\.");
-        
-        String body = new String(Base64.getUrlDecoder().decode(parts[1]), StandardCharsets.UTF_8);
-        ObjectMapper mapper = new ObjectMapper();
-        Map<String, Object> map = (Map<String, Object>) mapper.readValue(body, Map.class).get("data");
+
+        DecodedJWT decodedJWT = JWT.decode(jwt);
+        Map<String, Object> map = decodedJWT.getClaim("data").asMap();
```
jimmyjames: Not directly related to this change, but we should test with claims as they would be decoded normally by the library, instead of ad-hoc deserialization.

lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
```diff
@@ -507,7 +508,7 @@ public void shouldAcceptCustomMapClaimOfBasicObjectTypes() throws Exception {
         data.put("integer", 1);
         data.put("long", Long.MAX_VALUE);
         data.put("double", 123.456d);
-        data.put("date", new Date(123L));
+        data.put("date", new Date(123000));
```
jimmyjames: `java.util.Date` claims should be serialized to the seconds since the epoch. `Date` constructor takes a `long` representing the milliseconds since the epoch.

PR #422, opened 2020-05-27
add OSGi bundle support

hlavki:
### Changes
create OSGi bundle on build

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -1,7 +1,10 @@
+import aQute.bnd.gradle.Bundle
+
 plugins {
     id "com.jfrog.bintray" version "1.8.4"
     id "com.auth0.gradle.oss-library.java" version "0.11.0"
     id "jacoco"
+    id 'biz.aQute.bnd.builder'
 }
 
 logger.lifecycle("Using version ${version} for ${group}.${name}")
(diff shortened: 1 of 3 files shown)


lbalmaceda:
👋 we're not familiar with OSGi bundles. We use our own gradle [oss plugin](https://github.com/auth0/oss-library-gradle-plugin/) to prepare the jar and upload it to Bintray. 
What steps are involved in preparing such bundles and what's the output of this task?

hlavki:
This change affects only `META-INF/MANIFEST.MF` file, where osgi metadata are described.

lbalmaceda:
And what are the benefits of adding support for OSGi bundles?

hlavki:
the benefit is that you can use this library in osgi runtime. this MR defines dependencies and all needed osgi metadata.

lbalmaceda:
I'm not convinced we should accept this change. Especially because there's no one on our team who can troubleshoot this if we ran into any errors when packaging the next release. If this is added, looks like something we should be doing instead in our release plugin.

Recently we created an issue to track support for JDK 9 modules [here](https://github.com/auth0/oss-library-gradle-plugin/issues/19). If you want, you can create this issue and provide the details there.

jimmyjames:
We captured this issue in the OSS Plugin we use to package and deploy Java artifacts. The issue is https://github.com/auth0/oss-library-gradle-plugin/issues/20

You can follow along there; if you have a specific example you can share to demonstrate the issue and proposed fix, that would help us tremendously since we're not familiar with OSGI. Thanks!

Issue #419, opened 2020-04-30
Document thread-safety


qw3ry:
I would like to know whether or not the classes in this library are thread-safe.

I know there are already two closed issues asking for the thread safety of `JWTVerifier`.
I want to use an `Algorithm` with multiple threads for signing. Is this class threadsafe as well?

I recommend to document the thread-safety in the readme, or at least in the class / method javadocs.

lbalmaceda:
Hi @qw3ry. Thanks for raising this important topic for visibility. 
We agree the thread-safety should be documented carefully along the Javadoc and/or the readme. In the meantime until that analysis and documentation is done, here are some initial thoughts:
- The `Algorithm` class and subclasses are using class variables with the final modifier, and are also not mutating any received parameter. To my understanding, that alone helps a lot to reduce thread issues. 
- The `CryptoHelper` class makes use of `java.security` and `javax.crypto` classes. Nothing fancy, but also nothing unknown. There's no singleton there, every Algorithm instance would have its own helper.
The only "outsiders" could be:
a) the "apache-commons" dependency, used for b64 dec/enc operations. The class is thread-safe according to [their docs](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html).
b) the `KeyProvider` implementation of the asymmetric algorithms. What code goes in there is up to the developer. We only invoke that callback in order to get the public key given an ID.
And if anyone would like to contribute to this effort, we'd be happy to collaborate/review any PRs around this effort.

qw3ry:
[It looks like there are some issues with javax.crypto and thread safety](https://stackoverflow.com/a/31898893/2448440)

We chose to be better safe than sorry and create a new Algorithm instance each time, which is a lot better than debugging threading issues later.

lbalmaceda:
@qw3ry This is the result we've reached to: The library is almost completely thread-safe. We've pointed out the things we intend to change to take it to the next level ⚡ 

* There are a few references being kept as received instead of keeping a copy of them. Especially when dealing with arrays, lists, or maps. This can easily be fixed. Classes `PayloadDeserializer`, and `HSAlgorithm`
* The `Date` instance obtained from the clock is mutated. This supposes no problem for the default scenario that ships with this library, as the clock is meant to be customized for test purposes only. We will fix this by cloning the received value before mutating it. Classes `JWTVerifier`
[...]

PR #427, opened 2020-06-12
Fix and document thread-safety

lbalmaceda:
### Changes

Made a few fixes to behave better in terms of thread-safety. Also, documented the thread-safety of the classes after our evaluation. 

The reviewer might find useful to navigate the diff on a per-commit approach. The last commit is purely Javadoc changes.

### References

Closes https://github.com/auth0/java-jwt/issues/419

### Testing

I've added failure test cases right before the fix commit. Check the commit history.

### Checklist

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
+++ b/lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
@@ -6,11 +6,12 @@ import com.auth0.jwt.interfaces.DecodedJWT;
 import com.auth0.jwt.interfaces.ECDSAKeyProvider;
 import com.auth0.jwt.interfaces.RSAKeyProvider;
 
-import java.io.ByteArrayOutputStream;
 import java.security.interfaces.*;
 
 /**
  * The Algorithm class represents an algorithm to be used in the Signing or Verification process of a Token.
+ * <p>
+ * This class and its subclasses are thread-safe.
  */
 @SuppressWarnings("WeakerAccess")
 public abstract class Algorithm {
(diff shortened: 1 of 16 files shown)


lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
```diff
@@ -365,20 +366,20 @@ public String toString() {
     /**
      * Sign the given content using this Algorithm instance.
      *
-     * @param headerBytes an array of bytes representing the base64 encoded header content to be verified against the signature.
+     * @param headerBytes  an array of bytes representing the base64 encoded header content to be verified against the signature.
```
Widcket, line 369:
```suggestion
     * @param headerBytes an array of bytes representing the base64 encoded header content to be verified against the signature.
```

lib/src/test/java/com/auth0/jwt/algorithms/HMACAlgorithmTest.java
```diff
@@ -39,6 +40,18 @@ public void shouldGetStringBytes() throws Exception {
         assertTrue(Arrays.equals(expectedBytes, HMACAlgorithm.getSecretBytes(text)));
     }
 
+    @Test
+    public void shouldKeepCopyTheReceivedSecretArray() throws Exception {
```
Widcket, line 44:
```suggestion
    public void shouldCopyTheReceivedSecretArray() throws Exception {
```

lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
lbalmaceda: white spaces are being added because the class wasn't properly formatted following the project defaults (IDE default settings).

lib/src/test/java/com/auth0/jwt/algorithms/HMACAlgorithmTest.java
lbalmaceda: good catch!

PR #432, opened 2020-07-17
add many nullability annotations

thiakil:
### Changes

- No executable code changed
- Added `@Nullable` and `@NotNull` annotations to most public API methods (and some internal where nulls were checked for)
- Added `@Contact` to `com.auth0.jwt.algorithms.ECDSAAlgorithm#providerForKeys` as its null requirements are a little more complex
- Added `@Range` to `com.auth0.jwt.interfaces.Verification` leeway parameters

### References

Primarily useful for IDEs & static analysis, and also super helpful for Kotlin, where nullability is built into the type system.
This project is used in [Ktor](https://ktor.io/)

https://www.jetbrains.com/help/idea/nullable-and-notnull-annotations.html
https://kotlinlang.org/docs/reference/java-interop.html#nullability-annotations

### Testing

- [x] This change has been tested on the latest version of Java 

Gradle tests ran on OpenJDK 11.0.5

### Checklist

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -36,6 +36,7 @@ compileJava {
 dependencies {
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.10.3'
     implementation 'commons-codec:commons-codec:1.14'
+    implementation 'org.jetbrains:annotations:19.0.0'
     testImplementation 'org.bouncycastle:bcprov-jdk15on:1.60'
     testImplementation 'junit:junit:4.12'
     testImplementation 'net.jodah:concurrentunit:0.4.3'
--- a/lib/src/main/java/com/auth0/jwt/ClockImpl.java
+++ b/lib/src/main/java/com/auth0/jwt/ClockImpl.java
@@ -1,6 +1,7 @@
 package com.auth0.jwt;
 
 import com.auth0.jwt.interfaces.Clock;
+import org.jetbrains.annotations.NotNull;
 
 import java.util.Date;
 
@@ -17,6 +18,7 @@ final class ClockImpl implements Clock {
     ClockImpl() {
     }
 
+    @NotNull
     @Override
     public Date getToday() {
         return new Date();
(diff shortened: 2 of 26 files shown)


thibaultmeyer:
What hapoen to user not using IntelliJ ? Seems theses annotation need an extra dependency strongly linked to Jetbrains / IntelliJ

thiakil:
I'm not an Eclipse (or other tools) user, so I can't say for certain what it supports in terms of nullability annotations (I know it does support the `javax.annotations.*` variety in its static analysis).

It does not appear to be included in the maven-local pom file, and even if it does get pulled in by other projects, the jar is only 18kb and they are not runtime annotations so the jar is not required.

lbalmaceda:
I understand java 8 has [support for them](https://blogs.oracle.com/java-platform-group/java-8s-new-type-annotations), and given that we're already working on a v4 that requires JDK 8 as the minimum JDK, it could make sense to raise a PR against that branch. No additional dependencies should be required, and to my knowledge, these are just annotations that get processed at compile-time, not runtime. There shouldn't be any impact in terms of size/weight of the jar. 

@jimmyjames can you evaluate, please?

thiakil:
From your link:
> There is no set of default type annotations available out of the box in the Java SE 8 platform. All previous examples in this post used the Checker Framework.

As far as I know there is still no "standard" for these annotations - so it comes down to any tooling used (if any) supporting or being able to be configured to use the chosen annotations.

Kotlin (my primary motivation for the PR) supports a few variants (as listed in my linked section)

- JetBrains (`@Nullable` and `@NotNull` from the `org.jetbrains.annotations` package)
- Android (`com.android.annotations` and `android.support.annotations`)
- JSR-305 (`javax.annotation`, more details below)
- FindBugs (`edu.umd.cs.findbugs.annotations`)
- Eclipse (`org.eclipse.jdt.annotation`)
- Lombok (`lombok.NonNull`).

JSR-305 has _some_ Java 9 compatibility issues, though some quick Googling showed it may have been resolved early in the J9 cycle - as they're not from a core module, the `javax.annotation` package would cause failures as that package was claimed by a core module.

jimmyjames:
Thanks for providing the context and additional information @thiakil. As you mentioned, there are reported issues with the `javax.annotation` package and JDK9, though perhaps addressed now. That, along with the status of JSR-305 (dormant), as well as the unknown benefit or downsides to non-IntelliJ users, leads me to think this isn't something we'll incorporate in this SDK at this time. 

If this is a more pressing issue, you can open an issue and we can discuss further there, as well as get some more feedback from the community if this is something that is required to be addressed by this library.

PR #462, opened 2020-11-23
Thread-safe classes should be Shared statically

LeeHainie:
Thread-safe classes should be Shared statically,This improves performance and reduces class creation costs.


jimmyjames (approved):
Thanks @LeeHainie, this looks good 👍

PR #498, opened 2021-06-18
Expose the claim name in InvalidClaimException

ghostd:
### Changes

I need to know what claim is declared as invalid, so i propose to expose the claim name into the `InvalidClaimException`. Exposing the claim name to make checks is more reliable than compare the message exception.

### References

- support ticket
- community post
- StackOverflow post
- support forum thread

### Testing

Please describe how this can be tested by reviewers. Be specific about anything not tested and reasons why. If this library has unit and/or integration testing, tests should be added for new functionality and existing tests should complete without errors. 

- [*] This change adds test coverage
- [*] This change has been tested on the latest version of Java or why not

### Checklist

- [*] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
- [*] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
- [*] All existing and new tests complete without errors

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
@@ -328,13 +328,13 @@ public final class JWTVerifier implements com.auth0.jwt.interfaces.JWTVerifier {
                 assertValidAudienceClaim(jwt.getAudience(), (List<String>) expectedClaim.getValue(), false);
                 break;
             case PublicClaims.EXPIRES_AT:
-                assertValidDateClaim(jwt.getExpiresAt(), (Long) expectedClaim.getValue(), true);
+                assertValidDateClaim(jwt.getExpiresAt(), (Long) expectedClaim.getValue(), true, PublicClaims.EXPIRES_AT);
                 break;
             case PublicClaims.ISSUED_AT:
-                assertValidDateClaim(jwt.getIssuedAt(), (Long) expectedClaim.getValue(), false);
+                assertValidDateClaim(jwt.getIssuedAt(), (Long) expectedClaim.getValue(), false, PublicClaims.ISSUED_AT);
                 break;
             case PublicClaims.NOT_BEFORE:
-                assertValidDateClaim(jwt.getNotBefore(), (Long) expectedClaim.getValue(), false);
+                assertValidDateClaim(jwt.getNotBefore(), (Long) expectedClaim.getValue(), false, PublicClaims.NOT_BEFORE);
                 break;
             case PublicClaims.ISSUER:
                 assertValidIssuerClaim(jwt.getIssuer(), (List<String>) expectedClaim.getValue());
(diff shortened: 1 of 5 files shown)


lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -353,7 +353,7 @@ private void verifyClaimValues(DecodedJWT jwt, Map.Entry<String, Object> expecte
 
     private void assertClaimPresent(Claim claim, String claimName) {
         if (claim instanceof NullClaim) {
-            throw new InvalidClaimException(String.format("The Claim '%s' is not present in the JWT.", claimName));
+            throw new InvalidClaimException(String.format("The Claim '%s' is not present in the JWT.", claimName), claimName);
```
lbalmaceda, line 356: I like the goal that you are trying to achieve with this PR but I'm not convinced of usages like this one, in which the `claimName` value has to be passed twice in the same line. I assume the focus was on trying not to make any breaking changes.

Ideally, we would have two different exception classes. One that could represent a "missing claim" and one that could represent an "unexpected claim". Then those constructors would take just the claim name and the message would be set internally in there. I think this is something doable today also without breaking changes by subclassing the `InvalidClaimException`. Let me check with @jimmyjames what he thinks about adding two more exceptions to improve this use case.

ghostd: You're right. I tried to not break the current API. I can provide an other PR when you'll agree with a new API.

lbalmaceda:
@ghostd We met with @jimmyjames to discuss this feature. We agreed that we don't want the exception message/description to be part of our library's API contract, so that this PR would be a good addition. We've identified two scenarios that today are not completely covered in terms of exception handling when a claim is set to be expected in the verifier instance:

One of them is when an expected claim is not contained in the received payload JSON object. When that happens today, we throw `InvalidClaimException`. But that exception, as you point out, doesn't include the name of the missing claim. We suggest adding a new exception class that is a subclass of `InvalidClaimException` that would include the claim name, accessible through a public getter method. 

The other scenario we found is when an expected claim is present in the received payload JSON object, but its value is not the expected one. When that happens today, we throw `InvalidClaimException` without additional details. We suggest adding another subclass of `InvalidClaimException` that would include the claim name and the claim value, accessible through public getter methods.

By subclassing `InvalidClaimException`, devs can choose to continue to try/catching `InvalidClaimException` or capturing the more granular scenarios by try/catching the subclasses and accessing these informational properties. We wanted to know if these scenarios above would fit your use case or if we missed any detail. And of course, if you agree and can provide a PR with these changes. What are your thoughts?

ghostd:
My need is mainly the second scenario (i try to know which claim has an "invalid" value). Thanks for your time, i'll provide a PR.

ghostd:
I created the both exceptions (i can rename them i we find better names). For the claim value embedded into the `WrongClaimValueException`, i used the `toString` method because i didn't find how to extract a more precice value from a `Claim` instance.

lbalmaceda:
The diff tool check passed correctly but we would like to test this out locally before merging the PR. We're tracking this internally under `SDK-2836`. Cheers.

PR #501, opened 2021-06-29
Update OSS release plugin version

lbalmaceda:
### Changes

This PR updates the OSS release plugin version to 0.16.0, which makes use of a newer Javadoc HTML template.

I also removed the bintray plugin which is no longer used.

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -30,33 +30,13 @@ oss {
     }
 }
 
-def javaVersion = 8
-java {
-    toolchain {
-        languageVersion = JavaLanguageVersion.of(11)
-    }
-}
-
 compileJava {
     exclude 'module-info.java'
-
-    options.compilerArgs = ['--release', "$javaVersion"]
-}
-
-// Use latest JDK for javadoc generation
-tasks.withType(Javadoc).configureEach {
-    javadocTool = javaToolchains.javadocToolFor {
-        languageVersion = JavaLanguageVersion.of(16)
-    }
 }
 
 javadoc {
     // Exclude internal implementation package from javadoc
     excludes = ['com/auth0/jwt/impl', 'module-info.java']
-    // Specify Java version this project uses
-    // Otherwise would use version of javadoc toolchain by default which could
-    // cause compilation error mismatch between compileJava and javadoc creation
-    options.addStringOption('-release', "$javaVersion")
 }
 
 dependencies {
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -29,6 +29,12 @@ oss {
     }
 }
 
+java {
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(11)
+    }
+}
+
 compileJava {
     exclude 'module-info.java'
 }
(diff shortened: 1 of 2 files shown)


lib/build.gradle
```diff
@@ -30,33 +29,13 @@ oss {
...
-}
-
 compileJava {
     exclude 'module-info.java'
-
-    options.compilerArgs = ['--release', "$javaVersion"]
-}
-
-// Use latest JDK for javadoc generation
-tasks.withType(Javadoc).configureEach {
```
lbalmaceda, line 47: defined internally by the oss plugin

lib/build.gradle
```diff
@@ -30,33 +29,13 @@ oss {
...
 
-def javaVersion = 8
-java {
-    toolchain {
-        languageVersion = JavaLanguageVersion.of(11)
-    }
-}
-
 compileJava {
     exclude 'module-info.java'
```
lbalmaceda, line 33: this is the only line needed for module support to keep working, otherwise 
image

lib/build.gradle
```diff
@@ -30,33 +29,13 @@ oss {
...
-    }
 }
 
 javadoc {
     // Exclude internal implementation package from javadoc
     excludes = ['com/auth0/jwt/impl', 'module-info.java']
-    // Specify Java version this project uses
-    // Otherwise would use version of javadoc toolchain by default which could
-    // cause compilation error mismatch between compileJava and javadoc creation
-    options.addStringOption('-release', "$javaVersion")
```
lbalmaceda, line 59: defined internally by the oss plugin

lib/build.gradle
```diff
@@ -30,33 +29,13 @@ oss {
     }
 }
 
-def javaVersion = 8
-java {
-    toolchain {
-        languageVersion = JavaLanguageVersion.of(11)
```
lbalmaceda, line 36: Doesn't seem to be needed even when the library outputs a multi-jar with module support. 

image

lbalmaceda: @jimmyjames can you please double-check this one on your side? for me it keeps compiling after running something like `./gradlew clean build javadoc publishMavenJavaPublicationToMavenLocal --rerun-tasks`

jimmyjames: Yes, if you are running gradle with java 11 it will work. Using a toolchain you should be able to run gradle with say, java 8, and gradle will use the specified toolchain version for tasks. However the custom `compileModuleInfoJava` task does not appear to be respecting the toolchain, instead using the java version running gradle. Since our CI runs on 11 and this issue exists outside this change, I'd at least add a comment to the custom task with a todo to use the toolchain version so we can run the build with a different java version.

jimmyjames (changes requested):

lib/build.gradle
lbalmaceda: image

I tried changing my local JDK back to JDK 8 and ran the command above. The build is still successful. Do you think the comment is necessary? And, would that be for the "compileJava" task?

lbalmaceda: Per our discussion today, I'll add this block back to prevent potential issues when running the build using JDK 8.

jimmyjames (approved):

PR #512, opened 2021-09-03
[SDK-2751] Serialize audience claim when a List

### LINKED ISSUE #508: PublicClaims.AUDIENCE (aud) expected to be String or "Array of Strings", not Collection(aka List) by PayloadSerializer
### Describe the problem
By default jackson deserializes to ArrayList. Can not find an easy way to change this behavior.
so, having payload = `{exp=1630398915, aud=["firma_x","firma_y"]}` following code

```
    Map payloadMap = new ObjectMapper().readValue(payload, Map.class);
    String result = JWT.create()
                       .withPayload(payloadMap)
```
will cause an exception:

```
Caused by: java.lang.ClassCastException: class java.util.ArrayList cannot be cast to class [Ljava.lang.String; (java.util.ArrayList and [Ljava.lang.String; are in module java.base of loader 'bootstrap')
	at com.auth0.jwt.impl.PayloadSerializer.serialize(PayloadSerializer.java:40)
	at com.auth0.jwt.impl.PayloadSerializer.serialize(PayloadSerializer.java:18)
	at com.fasterxml.jackson.databind.ser.DefaultSerializerProvider._serialize(DefaultSerializerProvider.java:480)
```
Not sure, why special handling for PublicClaims.AUDIENCE is needed in first place, but it could be much more convenient  to support collections as well.

Note: no errors in other places(Claims) where collections are used.

### What was the expected behavior?

Collections can be handelt as well.

### Environment

- Version of this library used: 3.18.1
- Version of Java used: jdk-14.0.2
- Other modules/plugins/libraries that might be involved: com.fasterxml.jackson.core:jackson-databind:2.12.4

jimmyjames:
### Changes

As discussed in #508, when constructing a JWT using the `withPayload` method, the `aud` claim cannot be properly serialized, as we only expected it to be a String array (as it is when using the `withAudience` method). This change updates the `PayloadSerializer` to handle both String arrays and Lists to be serialized for the audience claim.

Fixes #508

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/impl/PayloadSerializer.java
+++ b/lib/src/main/java/com/auth0/jwt/impl/PayloadSerializer.java
@@ -5,8 +5,7 @@ import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.ser.std.StdSerializer;
 
 import java.io.IOException;
-import java.util.Date;
-import java.util.Map;
+import java.util.*;
 
 /**
  * Jackson serializer implementation for converting into JWT Payload parts.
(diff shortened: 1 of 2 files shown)


evansims (approved):

yeDor:
Hello @jimmyjames, @evansims, thanks for a fix. 
One question on this: why is implementation is aimed to List specifically, and not to the Collection, as it was discussed in https://github.com/auth0/java-jwt/issues/508 ?

Issue #507, opened 2021-08-17
3.8.0 introduced a breaking change


ebickle:
### Describe the problem
Version 3.8.0 of the library introduced a breaking change that broke binary compatibility. The method definition of `com.auth0.jwt.interfaces.Verification.withAudience` changed from `(String)` to `(String[])` when the parameter was modified from being a regular `String` to using a varargs.

The breaking change is not documented.

### What was the expected behavior?
Upgrading to a newer minor version of the library should not cause a breaking change in binary compatibility. Alternately, the breaking change should be documented.

The library should continue to have a proper `withAudience(String)` overload (or equivalent) to avoid the breaking change (assuming Java permits the overload alongside a varargs).

### Reproduction

N/A. Breaking change can be seen in source for PR #246 

### Environment

- **Version of this library used:** 3.18.1 (upgrading from 3.3.0)
- **Version of Java used:** Java 8 (OpenJDK variant)

jimmyjames:
Thanks @ebickle for raising this. You are correct that no breaking changes should be included in a minor release; and if it needs to be done for some reason, it should be documented. I suspect this one got through because it may _seem_ ok, but is indeed a binary incompatible change. We have added checks to our build pipeline to catch these sort of issues to prevent unintended breaking changes, but that was done after this change got through.

I'll see if there's a non-breaking way to restore the original method, though it may also be an issue as we'd need to add a new method to a public interface. If it's not possible, we'll retroactively update the changelog for 3.8.0 (obviously not ideal, we'll see what we can do).

jimmyjames:
@ebickle I assume you mean the change to `withIssuer`, not `withAudience`, as made in #288?

jimmyjames:
@ebickle I have a PR to restore `withIssuer(String issuer)` in #513. I'll keep it open for a couple days if you would like to review/try it. Thanks!

ebickle:
@jimmyjames Good catch, looks like I made a mistake in the original issue report. `withAudience` took a `String...` parameter in 3.3.0 and is the same in 3.8.0. `withIssuer` is the method that changed between the two.

Our build issue showed up because of some sort of unclean build or caching issue where a class was being deployed with a new library instead of being recompiled as it normally is. We've resolved that and aren't blocked on this issue any longer - but the PR will fix the binary compatibility for others, so probably worth doing regardless 😀

PR #513, opened 2021-09-03
[SDK-2758] Restore withIssuer

jimmyjames:
### Changes

#288 changed the signature of `withIssuer(String issuer)` to `withIssuer(String... issuer)`. As discussed in #507, this is a binary incompatible change. 

This change restores that binary compatibility, by introducing a new default method on the `Verification` interface, that simply delegates to the existing implementation. 

Notes:
* This change will cause the `apiDiff` check to fail, as the japicmp gradle plugin does not currently support configuring this check (see https://github.com/melix/japicmp-gradle-plugin/issues/43)
* Rather then configure the plugin to ignore the `Verification` interface, which would then never catch any issues, we'll need to bump the baseline compare version when this change is released.
* Regarding if adding a default method to an existing interface is binary compatible, [this documentation from Oracle](https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.5.6) provides a good explanation on why it should be considered binary compatible.
* No tests were added as part of this change,  as the `JWTVerifierTests` already contain a test that verify the behavior when calling `withIssuer` with a single string.
* When this change is released, we will need to update the `CHANGELOG` for 3.8.0 to indicate the binary incompatible change,  and note its fixed version.

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/interfaces/Verification.java
+++ b/lib/src/main/java/com/auth0/jwt/interfaces/Verification.java
@@ -8,6 +8,17 @@ import java.util.Date;
  * Holds the Claims and claim-based configurations required for a JWT to be considered valid.
  */
 public interface Verification {
+
+    /**
+     * Require a specific Issuer ("iss") claim.
+     *
+     * @param issuer the required Issuer value.
+     * @return this same Verification instance.
+     */
+    default Verification withIssuer(String issuer) {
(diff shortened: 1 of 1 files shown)


evansims (approved):

lbalmaceda (approved):
LGTM

ebickle (commented):
Looks good to me, thank you!

PR #530, opened 2022-02-28
Replace deprecated assertThat method in tests

poovamraj:
### Changes

The changes are for tests that used `org.junit.Assert.assertThat` method. This has been deprecated hence replaced with `org.hamcrest.MatcherAssert.assertThat`

### References
Java Doc of Deprecated Method - https://junit.org/junit4/javadoc/latest/org/junit/Assert.html#assertThat(java.lang.String,%20T,%20org.hamcrest.Matcher)

Screenshot 2022-02-28 at 16 03 52

### Testing
Since the changes are around test methods themselves, it is tested by running all the unit tests


jimmyjames (approved):
👍

PR #531, opened 2022-03-01
Remove apiDiff check for v4 development

jimmyjames:
This change removes the `apiDiff` check from v4 development, as there will be some binary/source breaking changes in the next major version. Once 4.0.0 is released, it will be added back to compare against 4.0.0

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,15 +23,15 @@ commands:
           name: Upload Coverage
           when: on_success
           command: bash <(curl -s https://codecov.io/bash) -Z -C $CIRCLE_SHA1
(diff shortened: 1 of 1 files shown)


poovamraj (approved):
👍 Looks good. We should probably add a task so that we don't forget to enable this. Also, how about just commenting it out and adding a hint saying that we should enable it after v4?

jimmyjames:
> We should probably add a task so that we don't forget to enable this. Also, how about just commenting it out and adding a hint saying that we should enable it after v4?

I updated the PR to comment it out, and created a GA task to re-enable.

poovamraj (approved):

PR #532, opened 2022-03-01
[SDK-3160] Replace com.auth0.jwt.interfaces.Clock with java.time.Clock

jimmyjames:
Replaces usages of `com.auth0.jwt.interfaces.Clock` with [java.time.Clock](https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html), which simplifies testing through the ability to inject a clock instance for testing.

Note that this is a breaking change, as the `JWTVerifier#build(com.auth0.jwt.interfaces.Clock clock)` method is public for testing purposes (the `Verification` interface does not expose this method, only `JWTVerifier#build()`). 

A separate PR will follow that removes the `Clock` interface and `ClockImpl` class, as they will no longer have any usages following this change.


poovamraj (approved):
LGTM 👍

PR #533, opened 2022-03-02
[SDK-3160] Remove internal Clock

jimmyjames:
As a follow-up to #532, this change removes the `Clock` and `ClockImpl` classes, as well as the tests. `java.time.Clock` should be used instead when needing to inject a clock for test purposes.


jimmyjames:
Codecov project check failure is due to removing code and tests.

poovamraj (approved):
👍

PR #535, opened 2022-03-02
Test only change - remove unnecessary throws clause from tests

jimmyjames:
Not much more description than in the title 😉 

Lots of changes, done through IDE inspection.


poovamraj (approved):
+1

PR #536, opened 2022-03-04
Testing Java LTS versions

poovamraj:
### Changes

- We have added 2 new tasks to run the Unit tests in different JVM LTS versions.
- Test class compilation will target Java 8 since we run unit tests in Java 8 (Previously this was compiled for Java 11)
- Jacoco `toolVersion` is explicitly mentioned since Java 17 is provided only above the specified version
- Had to ignore Tests for secp256k1 curve since it is disabled for Java 15+. We have to decide on whether we are going to support this and then decide on running these tests

### References
Jacoco Release note mentioning Java 17 support - https://www.jacoco.org/jacoco/trunk/doc/changes.html
Java dropping support for secp256k1 curve - https://www.oracle.com/java/technologies/javase/15-relnote-issues.html#JDK-8237219

### Testing
- Ensured the mentioned tasks are running in respective Java LTS versions
- Ensured the added tasks are run as part of the check task

### DIFF
--- a/lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
+++ b/lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
@@ -6,6 +6,7 @@ import com.auth0.jwt.interfaces.ECDSAKeyProvider;
 import com.auth0.jwt.interfaces.RSAKeyProvider;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
@@ -158,6 +159,7 @@ public class JWTCreatorTest {
     }
 
     @Test
+    @Ignore //todo: handle ecdsa curve secp256k1 disabled in Java 15+
     public void shouldAddKeyIdIfAvailableFromECDSAKAlgorithms() throws Exception {
         ECPrivateKey privateKey = (ECPrivateKey) PemUtils.readPrivateKeyFromFile(PRIVATE_KEY_FILE_EC_256K, "EC");
         ECDSAKeyProvider provider = mock(ECDSAKeyProvider.class);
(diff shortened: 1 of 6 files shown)


poovamraj:
CI Failure for Java 17 Unit Test task when testing unsupported curve for ECDSA 

Screenshot 2022-03-04 at 21 53 45

lib/src/test/java/com/auth0/jwt/JWTCreatorTest.java
```diff
@@ -6,6 +6,7 @@
 import com.auth0.jwt.interfaces.RSAKeyProvider;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
+import org.junit.Ignore;
```
jimmyjames, line 9: Can remove, not needed now

lib/src/test/java/com/auth0/jwt/JWTTest.java
```diff
@@ -4,6 +4,7 @@
 import com.auth0.jwt.interfaces.DecodedJWT;
 import org.hamcrest.collection.IsCollectionWithSize;
 import org.hamcrest.core.IsCollectionContaining;
+import org.junit.Ignore;
```
jimmyjames, line 7: Can remove, not needed now

lib/src/test/java/com/auth0/jwt/algorithms/ECDSAAlgorithmTest.java
```diff
@@ -6,6 +6,7 @@
 import com.auth0.jwt.interfaces.ECDSAKeyProvider;
 import org.hamcrest.Matchers;
 import org.hamcrest.collection.IsIn;
+import org.junit.Ignore;
```
jimmyjames, line 9: Can remove, not needed now.

jimmyjames (changes requested):
Looks good, just requested that you remove the unused `Ignore` import since we're excluding for java 17 only now.

jimmyjames (approved):

PR #537, opened 2022-03-07
[SDK-3149] Add Instant support

jimmyjames:
### Changes

**🕐 Adds support for `java.time.Instant` 🕐** 

This PR adds support for `java.time.Instant` for creating and verifying JWTs. Key points to note:
* `Instant` support is added to `Date` functionality; customers using `Date` can continue to do so.
* Additional methods added to interfaces have a default implementation using the underlying `Date` methods
* Internal implementation updated to use `Instant` for its improved functionality

**Note:** This PR does not include support for custom claim validation of type `Instant`. A separate PR will add support for this as well as address the current issue of `Date`-based claim validation not accounting for date/time claims being serialized as seconds since the epoch.

### References

https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html
### Testing

Existing unit tests provide good coverage, and tests added for new methods

PR #538, opened 2022-03-09
[SDK-3186] Support date/time custom claim validation

jimmyjames:
### Changes

Adds a new verification method to verify a custom date/time claim: `Verification#withClaim(String name, Instant value)`

Notes:
* Default implementation delegates to existing `withClaim(String name, Date value)` method to avoid breaking changes
* The current `Date`-based claim validation delegates to the new `Instant`-based validation, allowing us to remove `Date`-based validation code
* The current custom `Date`-based claim verification was arguably broken, or at least not intuitive. Because date claims are serialized as seconds since the epoch, any expected `Date` claim would need to have its milliseconds set to zero to be considered equal. This change fixes that by truncating the expected claim's value to seconds.

### References

Builds on #537

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
@@ -174,8 +174,14 @@ public final class JWTVerifier implements com.auth0.jwt.interfaces.JWTVerifier {
 
         @Override
         public Verification withClaim(String name, Date value) throws IllegalArgumentException {
+            return withClaim(name, value != null ? value.toInstant() : null);
+        }
+
+        @Override
+        public Verification withClaim(String name, Instant value) throws IllegalArgumentException {
             assertNonNull(name);
-            requireClaim(name, value);
+            // Since date-time claims are serialized as epoch seconds, we need to compare them with only seconds-granularity
+            requireClaim(name, value != null ? value.truncatedTo(ChronoUnit.SECONDS) : null);
             return this;
         }
 
@@ -371,8 +377,8 @@ public final class JWTVerifier implements com.auth0.jwt.interfaces.JWTVerifier {
             isValid = value.equals(claim.asBoolean());
         } else if (value instanceof Double) {
             isValid = value.equals(claim.asDouble());
-        } else if (value instanceof Date) {
-            isValid = value.equals(claim.asDate());
+        } else if (value instanceof Instant) {
+            isValid = value.equals(claim.asInstant());
         } else if (value instanceof Object[]) {
             List<Object> claimArr;
             Object[] claimAsObject = claim.as(Object[].class);
--- a/lib/src/main/java/com/auth0/jwt/interfaces/Verification.java
+++ b/lib/src/main/java/com/auth0/jwt/interfaces/Verification.java
@@ -178,7 +178,7 @@ public interface Verification {
 
     /**
      * Require a specific Claim value. Note that date-time claims are serialized as seconds since the epoch; when verifying
-     * date-time claim value, any time units more granular than seconds will be discarded.
+     * date-time claim value, any time units more granular than seconds will not be considered.
      *
      * @param name  the Claim's name.
      * @param value the Claim's value.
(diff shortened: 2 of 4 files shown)


lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
```diff
@@ -517,7 +517,7 @@ public void shouldValidateCustomClaimOfTypeBoolean() {
     @Test
     public void shouldValidateCustomClaimOfTypeDate() {
         String token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYW1lIjoxNDc4ODkxNTIxfQ.mhioumeok8fghQEhTKF3QtQAksSvZ_9wIhJmgZLhJ6c";
-        Date date = new Date(1478891521000L);
+        Date date = new Date(1478891521123L);
```
jimmyjames, line 520: Changing the date time to include milliseconds would cause this test to fail without the change to truncate milliseconds.

jimmyjames:
Codecov failing because default implementation isn't covered. I'll look into adding a test for that (and other default implementations in that interface).

jimmyjames:
> Codecov failing because default implementation isn't covered. I'll look into adding a test for that (and other default implementations in that interface).

Additional tests added for default method implementations in `Verification` instance, as well as the `null` handling of `Verification` implementation methods.

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -371,8 +377,8 @@ private void assertValidClaim(Claim claim, String claimName, Object value) {
             isValid = value.equals(claim.asBoolean());
         } else if (value instanceof Double) {
             isValid = value.equals(claim.asDouble());
-        } else if (value instanceof Date) {
```
jimmyjames, line 374: since we delegate custom `Date` expectations to `withClaim(String name, Instant value)`, we can remove `Date`-based check.

lib/src/test/java/com/auth0/jwt/interfaces/VerificationTest.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasEntry;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.junit.Assert.assertThrows;
+
+/**
+ * Tests for any default method implementations in the {@link Verification} interface.
```
poovamraj, line 17: I found that we have a default method that we have previously added in the Verification interface, since the point of default methods in the interface is to avoid breaking changes, and we are releasing a major that can have breaking changes, why not remove the default method and implement it the right way? 

This will break one method implementation if the user is implementing the Verification interface but I guess it should be fine. Your thoughts on this @jimmyjames

lib/src/main/java/com/auth0/jwt/interfaces/Verification.java
```diff
@@ -176,7 +177,8 @@ default Verification withAnyOfAudience(String... audience) {
     Verification withClaim(String name, String value) throws IllegalArgumentException;
 
     /**
-     * Require a specific Claim value.
+     * Require a specific Claim value. Note that date-time claims are serialized as seconds since the epoch; when verifying
+     * date-time claim value, any time units more granular than seconds will be discarded.
```
poovamraj, line 181: Could we rather say "not be considered/included" instead of "discarded"? Since I already have the context of the fix I am able to understand this easily. Not sure for the first time reader. But I could be wrong here

poovamraj, line 181: Could we rather say "not be considered/included" instead of "discarded"? Since I already have the context of the fix I am able to understand this easily. Not sure for the first time reader. But I could be wrong here

poovamraj (commented):
Hey Jim, left a couple of subjective points on the PR. let me know what you think of it

lib/src/main/java/com/auth0/jwt/interfaces/Verification.java
jimmyjames: Updated the PR to use the terminology "will not be considered"

lib/src/test/java/com/auth0/jwt/interfaces/VerificationTest.java
jimmyjames: Removing the default methods that exist in v3 makes sense. And while we could add a new method to the interface without a default implementation in a new major, if there's a reasonable default implementation to make upgrading versions easier, I prefer that. We can remove the default method implementations for existing default methods in a separate PR.

poovamraj (approved):

PR #541, opened 2022-03-11
[SDK-3192] Deprecate secp256k1 curve for EC Algorithms and undeprecate single key constructors & signing methods

poovamraj:
Changes
We are deprecating the secp256k1 curve since Java 15+ has disabled this. This algorithm "ES256K" is not mentioned in RFC and removal of this algorithm in future is planned.

We are also removing deprecation notice for constructing Algorithm using a single key and signing contents using single parameter instead since we have found these to be useful methods

Testing
No tests since just a documentation change

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
+++ b/lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
@@ -182,8 +182,10 @@ public abstract class Algorithm {
      *
      * @param keyProvider the provider of the Public Key and Private Key for the verify and signing instance.
      * @return a valid ECDSA256 Algorithm.
-     * @deprecated This method is deprecated since Java 15+ disables SECP-256K1 Curve. Using this in Java 15+ will throw a SignatureException
      * @throws IllegalArgumentException if the Key Provider is null.
+     * @deprecated The SECP-256K1 Curve algorithm has been disabled beginning in Java 15.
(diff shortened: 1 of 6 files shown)


lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
```diff
@@ -182,8 +182,10 @@ public static Algorithm HMAC256(byte[] secret) throws IllegalArgumentException {
      *
      * @param keyProvider the provider of the Public Key and Private Key for the verify and signing instance.
      * @return a valid ECDSA256 Algorithm.
+     * @deprecated This method is deprecated since Java 15+ disables SECP-256K1 Curve. Using this in Java 15+ will throw a SignatureException
```
jimmyjames, line 185: How about:

"The SECP-256K1 Curve algorithm has been disabled beginning in Java 15, and removed in Java 16+. Use of this method in those unsupported Java versions will throw a {@link SignatureException}. This method will be removed in the next major version. See <a href="https://bugs.openjdk.java.net/browse/JDK-8251547"></a> for additional information"

jimmyjames (changes requested):
Proposed an updated deprecation description with additional info. Also check if the README also requires an update, thanks!

poovamraj:
Hey @jimmyjames 👋 Considering we are releasing the version bump this week, hope we could merge these changes along with them and release them together

PR #549, opened 2022-03-15
[SDK-3171] Fix header claims serialization

jimmyjames:
### Changes

Currently, we serialize the header claims using Jackson's default `writeString`. This causes date-time based claims to be serialized as milliseconds since the epoch, instead of the required seconds since the epoch.

This PR refactors much of the logic of the `PayloadSerializer` out to a new class `ClaimsSerializer`, which allows a subclass to override how the claim is written (needed to handle the `aud` claim serialization for the payload).

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/impl/ClaimsSerializer.java
+++ b/lib/src/main/java/com/auth0/jwt/impl/ClaimsSerializer.java
@@ -24,7 +24,6 @@ public class ClaimsSerializer<T extends ClaimsHolder> extends StdSerializer<T> {
     @Override
     public void serialize(T holder, JsonGenerator gen, SerializerProvider provider) throws IOException {
         gen.writeStartObject();
-        holder.getClaims().forEach((key, val) -> System.out.println(key + ": " + val));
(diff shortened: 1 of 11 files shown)


lib/src/main/java/com/auth0/jwt/impl/ClaimsSerializer.java
```diff
@@ -0,0 +1,87 @@
...
+public class ClaimsSerializer<T extends ClaimsHolder> extends StdSerializer<T> {
+
+    public ClaimsSerializer(Class<T> t) {
+        super(t);
+    }
+
+    @Override
+    public void serialize(T holder, JsonGenerator gen, SerializerProvider provider) throws IOException {
+        gen.writeStartObject();
+        holder.getClaims().forEach((key, val) -> System.out.println(key + ": " + val));
```
poovamraj, line 27: I guess we need to remove this

jimmyjames: good catch, new commit added to remove

PR #560, opened 2022-03-22
Using Predicates for Verification

poovamraj:
Context
- Previously all our verification happened depending on the type of the claim value. This can be seen in the [verifyClaimValues](https://github.com/auth0/java-jwt/blob/master/lib/src/main/java/com/auth0/jwt/JWTVerifier.java#L320) method with so many `instanceof` checks.
- This caused us to implement unwanted hacks like `NonEmptyClaim` type to do certain validations
- To avoid this, we are moving all our claim validations to a collection of predicates that can be built.

Changes
- All assertion calls (assertValidInstantClaim, assertValidAudienceClaim) are moved into BaseVerification
- Asserting Collections (old code - [Link](https://github.com/auth0/java-jwt/blob/master/lib/src/main/java/com/auth0/jwt/JWTVerifier.java#L375)) now has its own method `assertValidCollectionClaim`
- `requireClaim` method is renamed to `checkIfNeedToRemove` to resemble its functionality which is to remove validation for any claim for which the value is passed as null
- We have rearranged the position of the methods for better readability. This causes the diff to be really tough to compare with.
- Introduced `getLeewayFor` method which will be used get leeway for each claim.
- `customLeeways` will now store the custom leeway value provided for corresponding Claims. This avoids the old behavior where leeway was stored as a claim value to be checked.
- We will no longer need to store Generic Objects in the `expectClaim` member
- Hack used to compare `aud` claim using [AUDIENCE_EXACT](https://github.com/auth0/java-jwt/blob/master/lib/src/main/java/com/auth0/jwt/JWTVerifier.java#L28) and [AUDIENCE_ANY](https://github.com/auth0/java-jwt/blob/master/lib/src/main/java/com/auth0/jwt/JWTVerifier.java#L29) has now been removed. Since same claim name is used for key, Map ensures uniqueness mentioned in the [document](https://github.com/auth0/java-jwt/blob/master/lib/src/main/java/com/auth0/jwt/interfaces/Verification.java#L42)
- `NonEmptyClaim` class removed

Testing
- We made few fixes to existing test cases to make it compatible to our new changes
- We haven't added more tests since this was only an internal refactor and no behaviour has been added or modified to be tested

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
@@ -24,17 +24,12 @@ import java.util.function.BiPredicate;
 @SuppressWarnings("WeakerAccess")
 public final class JWTVerifier implements com.auth0.jwt.interfaces.JWTVerifier {
     private final Algorithm algorithm;
-    final Map<String, Object> expectedClaims;
-    private final Clock clock;
+    final Map<String, BiPredicate<Claim, DecodedJWT>> expectedChecks;
     private final JWTParser parser;
 
-    static final String AUDIENCE_EXACT = "AUDIENCE_EXACT";
-    static final String AUDIENCE_CONTAINS = "AUDIENCE_CONTAINS";
-
-    JWTVerifier(Algorithm algorithm, Map<String, Object> expectedClaims, Clock clock) {
+    JWTVerifier(Algorithm algorithm, Map<String, BiPredicate<Claim, DecodedJWT>> expectedClaims) {
         this.algorithm = algorithm;
-        this.expectedClaims = Collections.unmodifiableMap(expectedClaims);
-        this.clock = clock;
+        this.expectedChecks = Collections.unmodifiableMap(expectedClaims);
         this.parser = new JWTParser();
     }
 
@@ -51,9 +46,11 @@ public final class JWTVerifier implements com.auth0.jwt.interfaces.JWTVerifier {
 
     public static class BaseVerification implements Verification {
         private final Algorithm algorithm;
-        private final Map<String, Object> expectedClaims;
+        private final Map<String, BiPredicate<Claim, DecodedJWT>> expectedChecks;
         private long defaultLeeway;
+        private final Map<String, Long> customLeeways;
         private boolean ignoreIssuedAt;
+        private Clock clock;
 
         BaseVerification(Algorithm algorithm) throws IllegalArgumentException {
             if (algorithm == null) {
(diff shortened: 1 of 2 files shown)


poovamraj:
This is a seperate PR into the existing predicates support branch, so as to ensure the unrefactored version is left clean for us to make flexible decisions

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -51,43 +46,54 @@ static Verification init(Algorithm algorithm) throws IllegalArgumentException {
...
+                }
+                return true;
+            }));
             return this;
         }
 
         @Override
         public Verification withSubject(String subject) {
-            requireClaim(PublicClaims.SUBJECT, subject);
+            checkIfNeedToRemove(PublicClaims.SUBJECT, subject, (claim, decodedJWT) -> subject.equals(claim.asString()));
```
panga, line 80: This code will fail with NPE if null is passed as subject. Is it the desired behavior?

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -51,43 +46,54 @@ static Verification init(Algorithm algorithm) throws IllegalArgumentException {
...
+        private Clock clock;
 
         BaseVerification(Algorithm algorithm) throws IllegalArgumentException {
             if (algorithm == null) {
                 throw new IllegalArgumentException("The Algorithm cannot be null.");
             }
 
             this.algorithm = algorithm;
-            this.expectedClaims = new HashMap<>();
+            this.expectedChecks = new HashMap<>();
```
panga, line 61: What do you think of using LinkedHashMap to maintain the insertion order of claims? It would make checks run consistently in the order they are declared in the code.

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
```diff
@@ -453,10 +418,10 @@ public void shouldThrowOnInvalidCustomClaimValue() {
         exception.expect(InvalidClaimException.class);
         exception.expectMessage("The Claim 'name' value doesn't match the required one.");
         String token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJuYW1lIjpbInNvbWV0aGluZyJdfQ.3ENLez6tU_fG0SVFrGmISltZPiXLSHaz_dyn-XFTEGQ";
-        Map<String, Object> map = new HashMap<>();
-        map.put("name", new Object());
-        JWTVerifier verifier = new JWTVerifier(Algorithm.HMAC256("secret"), map, Clock.systemUTC());
-        verifier.verify(token);
+        JWTVerifier.init(Algorithm.HMAC256("secret"))
+                .withClaim("name", "")
```
panga, line 422: Why this parameter was changed from Object to empty string?

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -127,49 +133,54 @@ public Verification ignoreIssuedAt() {
...
+            checkIfNeedToRemove(PublicClaims.JWT_ID, jwtId, ((claim, decodedJWT) -> jwtId.equals(claim.asString())));
             return this;
         }
 
         @Override
         public Verification withClaimPresence(String name) throws IllegalArgumentException {
             assertNonNull(name);
-            requireClaim(name, NonEmptyClaim.getInstance());
+            withClaim(name, ((claim, decodedJWT) -> {
+                if (claim instanceof NullClaim) {
```
panga, line 144: Can't we just use `claim.isNull()` ?

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -24,17 +24,12 @@
...
 
-    static final String AUDIENCE_EXACT = "AUDIENCE_EXACT";
-    static final String AUDIENCE_CONTAINS = "AUDIENCE_CONTAINS";
-
-    JWTVerifier(Algorithm algorithm, Map<String, Object> expectedClaims, Clock clock) {
+    JWTVerifier(Algorithm algorithm, Map<String, BiPredicate<Claim, DecodedJWT>> expectedClaims) {
         this.algorithm = algorithm;
-        this.expectedClaims = Collections.unmodifiableMap(expectedClaims);
-        this.clock = clock;
+        this.expectedChecks = Collections.unmodifiableMap(expectedClaims);
```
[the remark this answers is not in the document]
poovamraj: Yes exactly! Nice catch, I will change the constructor name.

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
poovamraj: Since `checkIfNeedToRemove` will check for `null` value for the parameter `subject` and remove the check if it is null, I guess there shouldn't be an NPE issue here

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
poovamraj: Nice catch @panga! I was actually thinking of it, and in future, we can have multiple checks for a single claim by using an ArrayList instead of a Map. but since that is a behaviour change, I will change this one to LinkedHashMap

lib/src/test/java/com/auth0/jwt/JWTVerifierTest.java
poovamraj: We don't have a constructor that accepts an object. This test is to ensure that a claim in a given name is not present and not about the value, I decided to go ahead with an empty string, I will add some value instead of an empty string here.

lib/src/main/java/com/auth0/jwt/JWTVerifier.java
poovamraj: `withClaimPresence` method is used to check whether a claim in the given name is present.  Our next task will be handling `null` values in our JWT (SDK-3158). Since in that task we will be acknowledging the Claim's presence in JWT (i.e not throw Exception as in here) this method's behaviour will change in that task. I guess it would be better for us to handle it there.

poovamraj:
Going ahead with the merge here. I have addressed the reviews mentioned here. Since another PR will be created to merge into v4-dev, We can go through this again if needed

PR #561, opened 2022-03-23
Add lint checks

poovamraj:
### Context
We have added Lint checks to ensure consistent coding style across our repository

### Changes

- Made necessary changes to provide consistent code format
- We have used Google code style with little modifications to suit our style and need
- We have added avoiding System.out.println as a seperate rule
- We have made all the changes requested in linter.
- Any error in this check will throw an error in our CI with details in the console 

### References
https://checkstyle.sourceforge.io/

### Testing
- Tested whether the failure of Lint check fails the CI
- Existing test cases pass
- No change in code, just changes in formatting

### DIFF
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -7,6 +7,7 @@ plugins {
 
 checkstyle {
     toolVersion '10.0'
+    checkstyleTest.enabled = false //We are disabling lint checks for tests
 }
 
 logger.lifecycle("Using version ${version} for ${group}.${name}")
--- a/lib/build.gradle
+++ b/lib/build.gradle
@@ -2,6 +2,11 @@ plugins {
     id 'java'
     id 'jacoco'
     id 'com.auth0.gradle.oss-library.java'
+    id 'checkstyle'
+}
+
+checkstyle {
+    toolVersion '10.0'
 }
 
 logger.lifecycle("Using version ${version} for ${group}.${name}")
--- a/lib/src/main/java/com/auth0/jwt/impl/PublicClaims.java
+++ b/lib/src/main/java/com/auth0/jwt/impl/PublicClaims.java
@@ -1,6 +1,8 @@
 package com.auth0.jwt.impl;
 
-
+/**
+ * Contains the claim name for all Public claims.
+ */
 public interface PublicClaims {
 
     //Header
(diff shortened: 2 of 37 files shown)


lib/src/main/java/com/auth0/jwt/impl/PublicClaims.java
```diff
@@ -1,6 +1,8 @@
 package com.auth0.jwt.impl;
 
-
+/**
+ * Contains the claim name for all Public claims.
+ */
 public interface PublicClaims {
```
panga, line 6: Any reason for this class to be in the `impl` (internal package) ?
I see some other classes using `public` in this package, this is something to consider change in future.
This package should also be removed from module-info (java 9+).

lib/build.gradle
```diff
@@ -2,6 +2,12 @@ plugins {
     id 'java'
     id 'jacoco'
     id 'com.auth0.gradle.oss-library.java'
+    id 'checkstyle'
+}
+
+checkstyle {
+    toolVersion '10.0'
+    checkstyleTest.enabled = false //We are disabling lint checks for tests
```
panga, line 10: Why disabling tests? what's the side-effect on enabling there?

lib/src/main/java/com/auth0/jwt/impl/PublicClaims.java
poovamraj: My assumption on this class being internal is to provide less API surface area for the consumer. But as you mentioned having it as public negates the effort and we don't have any other option currently since the package structure is that way.

For Java 9+, if the project is using the module feature, then they won't be able to import this package as it can be seen here - https://github.com/auth0/java-jwt/blob/v4-dev/lib/src/main/java/module-info.java.

lib/build.gradle
poovamraj: We are disabling this for tests since there 500+ errors in our test code base as well. I thought we could refactor it in a later release. What would be your suggestion?

panga: Makes sense 👍

lib/src/main/java/com/auth0/jwt/impl/PublicClaims.java
panga: The next major version is a big opportunity to remove public from interfaces/classes in impl package.
Module system is under utilized in Java, even if you using latest Java 18 everyone still relying on classloader.
Lets make sure this get reviewed before release, ideally should have none.

panga (approved):

lib/src/main/java/com/auth0/jwt/impl/PublicClaims.java
poovamraj: I agree @panga, I have created an internal ticket [SDK-3226] to assigned it to this major. Since @jimmyjames has more context over this library, we should decide on this once he is back.

PR #570, opened 2022-03-31
Improve keyprovider reliability

poovamraj:
### Changes
- Added PrivateKeyDetail class
- Deprecated getPrivateKey and getPrivateKeyId
- Default Implementation for getPrivateKeyDetails
- Getting the Private Key at the point of signing and passing it as parameters so that another getPrivateKeyDetails which can change the Key details

### References
https://github.com/auth0/java-jwt/issues/503
https://github.com/auth0/java-jwt/pull/517

### Testing
Added a Key Rotation Test (Proposed in #517). This test takes a long time to complete which increases the CI time.
Fixed existing mocks

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
+++ b/lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
@@ -4,8 +4,10 @@ import com.auth0.jwt.exceptions.SignatureGenerationException;
 import com.auth0.jwt.exceptions.SignatureVerificationException;
 import com.auth0.jwt.interfaces.DecodedJWT;
 import com.auth0.jwt.interfaces.ECDSAKeyProvider;
+import com.auth0.jwt.interfaces.PrivateKeyDetail;
 import com.auth0.jwt.interfaces.RSAKeyProvider;
 
+import java.security.PrivateKey;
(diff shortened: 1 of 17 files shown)


lib/src/main/java/com/auth0/jwt/algorithms/Algorithm.java
```diff
@@ -367,14 +369,32 @@ public String toString() {
...
+     * @param headerBytes  an array of bytes representing the base64 encoded header content
+     *                     to be verified against the signature.
+     * @param payloadBytes an array of bytes representing the base64 encoded payload content
+     *                     to be verified against the signature.
+     * @param privateKey   A private key with which the content has to be signed
+     *
+     * @return the signature in a base64 encoded array of bytes
+     * @throws SignatureGenerationException if the Key is invalid.
+     */
+    public byte[] sign(byte[] headerBytes, byte[] payloadBytes, PrivateKey privateKey)
```
poovamraj, line 388: We are exposing this new method in Algorithm class which takes in a Private Key, this is to ensure the same Private key used to build the header claim is used for signing. But this cannot be achieved without making this method public. This takes in a Private Key where Algorithm already has a private key which is contradictory

PR #572, opened 2022-04-05
Replace expected exception with assertThrows

poovamraj:
### Changes
To check for the values that are present in exception we used custom matchers. We have instead replaced them with `assertThrows`

### Testing
Code Refactor on existing tests


jimmyjames (commented):
Changes look good. We should understand why the codecov check is failing and fix that.

Issue #558, opened 2022-03-21
undesired behavior for withClaimPresence / alternatively inadequate documentation


Redeeman:
(i do not agree to the code of conduct)

### Describe the problem

Implementing JWT in an application, I decided that I did not want to accept tokens without expiry, so I activated .withClaimPresence( "exp" ), to ensure that any token without expiration would not be accepted. This appears to have disabled verification of the expiry.

All the docs say for this method is "Require a claim to be present, with any value.", which to be fair, is what is happening, but since there is no other explicit configuration for expiry beyond the leeway settings, I had assumed the automatic verification of expiry would continue.

I would at the very least recommend putting a stern warning in the documentation for "withClaimPresence()" about this behavior. It might however be better to simply change the behavior to always process expiry if the claim exists. Functionality already exists to functionally disable validation if wanted, by for example setting the leeway extremely high (though I do not see why someone might want to do that).

Alternatively perhaps explicit control to enable/disable verification of expiry.

Let me know what you think, example:

Algorithm algorithm = Algorithm.HMAC256( REDACTED );
JWTVerifier verifier = JWT.require( algorithm )
        .withClaimPresence( "sub" )
        .withClaimPresence( "iss" )
        .withClaimPresence( "exp" ) <-- this must not be activated, or it will NOT validate expiry
        .acceptLeeway( 5 )
        .withClaim( "pmtyp", "user" )
        .build();

DecodedJWT jwt = verifier.verify( token );

### What was the expected behavior?

that it validate the expiry even though .withClaimPresence( "exp" ) was called

### Reproduction
run snippet

poovamraj:
Hi @Redeeman, we tried to reproduce the issue using the following code and it doesn't seem to reproduce for us. Every time the expiry is validated even if presence is checked for it. Can you provide a full sample where the issue reproduces?

```java
Algorithm algorithm = Algorithm.HMAC256( "REDACTED" );
            JWTVerifier verifier = JWT.require( algorithm )
                    .withClaimPresence( "sub" )
                    .withClaimPresence( "exp" )
                    .acceptLeeway( 5 )
                    .build();

[...]

PR #573, opened 2022-04-06
[SDK-3231] Added support for multiple checks on a single claim

poovamraj:
### Changes
- We have added option where multiple claim checks can be done for same claim. 
- This causes breaking change to the existing behaviour where only one check can be done over a claim.
- This also overrides the behaviour where if null is passed to a claim check, the check will get removed. Instead, now it will check for null value for that claim

**Before** this code removes the check for `claim` with `123`
**After** the code checks for claim to have value `123` and then checks for value `null`
```java
JWT.require(Algorithm.HMAC256("secret"))
                .withClaim("claim", 123)
                .withClaim("claim", (Integer) null)
                .build();
```

**Before** this code overrides the check for `claim` with `check1` to `check2`
**After** the code checks for `claim` to have value `check1` and `check2`
```java
JWT.require(Algorithm.HMAC256("secret"))
                .withClaim("claim", "check1")
                .withClaim("claim", "check2")
                .build();
```

**Before** this code doesn't check for the `claim` value
**After** the code checks for `claim` to have `null`
```java
JWT.require(Algorithm.HMAC256("secret"))
                .withClaim("claim", (Integer) null)
                .build();
```

### References
#558 

### Testing
- We have added tests to ensure null values are checked properly in the claim
- We have tested that multiple claim checks are added
- Improved branch coverage to ~98%

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -7,7 +7,7 @@ coverage:
       default:
         if_no_uploads: success
         if_ci_failed: error
-    changes: false
+    changes: true
     project:
       default:
         target: auto
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -13,3 +13,5 @@ coverage:
         target: auto
         if_no_uploads: error
 comment: false
+codecov:
+  allow_coverage_offsets: True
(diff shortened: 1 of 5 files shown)


jimmyjames (approved):
👍

PR #574, opened 2022-04-12
[SDK-3226] Expose claim header constants

jimmyjames:
Removes the `PublicClaims` interface that held constants for claim and header names. Replaces with new `StandardClaims` and `HeaderParams` classes to hold these constant values, and are now exported as part of the public API.


poovamraj (approved):
👍  LGTM

PR #577, opened 2022-04-20
[SDK-3159] JavaDoc updated

poovamraj:
### Changes
We have updated the Java doc with relevant information and made a few required changes.

This includes 
- removing deprecation notices
- using more Java doc features
- other grammatical improvements

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -19,6 +19,8 @@ import java.util.Map.Entry;
  * from a given Header and Payload content.
  * <p>
  * This class is thread-safe.
+ *
+ * TODO Poovam - Should we claim thread safety since KeyProvider implementation can cause signing differ
  */
 @SuppressWarnings("WeakerAccess")
 public final class JWTCreator {
--- a/lib/src/main/java/com/auth0/jwt/JWTCreator.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTCreator.java
@@ -19,8 +19,6 @@ import java.util.Map.Entry;
  * from a given Header and Payload content.
  * <p>
  * This class is thread-safe.
- *
- * TODO Poovam - Should we claim thread safety since KeyProvider implementation can cause signing differ
  */
 @SuppressWarnings("WeakerAccess")
 public final class JWTCreator {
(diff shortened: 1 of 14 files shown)


poovamraj:
Project coverage has decreased by 0.01% since we removed `withAnyOfAudienceDeafultImplShouldThrow` to test default method which has been removed

lib/src/main/java/com/auth0/jwt/JWTCreator.java
```diff
@@ -19,6 +19,8 @@
  * from a given Header and Payload content.
  * <p>
  * This class is thread-safe.
+ *
+ * TODO Poovam - Should we claim thread safety since KeyProvider implementation can cause signing differ
```
jimmyjames, line 23: Do you mean that an implementation of KeyProvider may not be thread-safe, rendering this class not thread-safe? If so, we can claim thread safety of this class, and also document that implementations of KeyProvider must be thread-safe.

lib/src/main/java/com/auth0/jwt/RegisteredClaims.java
```diff
@@ -11,37 +11,44 @@ private RegisteredClaims() {
...
 
     /**
      * The "iat" (issued at) claim identifies the time at which the JWT was issued.
+     * Refer RFC 7529 <a href="https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.6">Section 4.1.6</a>
      */
     public static String ISSUED_AT = "iat";
 
     /**
      * The "jti" (JWT ID) claim provides a unique identifier for the JWT.
+     * Refer RFC 7529 <a href="https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7">Section 4.1.7</a>
```
jimmyjames, line 51: 👍 these are nice

lib/src/main/java/com/auth0/jwt/interfaces/Verification.java
```diff
@@ -22,52 +23,40 @@ default Verification withIssuer(String issuer) {
...
+     * Verifies whether the JWT contains an Audience ("aud") claim contain at least one of the specified audiences.
+     * This check is case-sensitive. An empty array is considered as a {@code null}.
      *
      * @param audience the required Audience value for which the "aud" claim must contain at least one value.
      * @return this same Verification instance.
      */
-    default Verification withAnyOfAudience(String... audience) {
-        throw new UnsupportedOperationException("withAnyOfAudience");
-    }
+    Verification withAnyOfAudience(String... audience);
```
jimmyjames, line 59: 👍 thanks for removing the default implementation

jimmyjames (commented):
Looks great - I left a comment regarding your TODO. Once we address that, this is good to go 👍

jimmyjames (approved):

PR #579, opened 2022-05-03
[SDK-3311] Added protection against CVE-2022-21449

poovamraj:
### Changes
This PR will implement defence against CVE-2022-21449 by validating the signatures structure so that
- The entire signature is not filled with zero
- The value of R and S are not zero
- The value of R and S are not less than 1 compared to the ECDSA Algorithm's order

### References
- [CVE-2022-21449](https://nvd.nist.gov/vuln/detail/CVE-2022-21449)

### Testing
We have added Unit Tests to check for the vulnerable signature. We have ensured the logic we use to validate the signature is also unit tested

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@
 
 A Java implementation of [JSON Web Token (JWT) - RFC 7519](https://tools.ietf.org/html/rfc7519).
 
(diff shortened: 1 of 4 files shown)


lbalmaceda:
LGTM

PR #583, opened 2022-05-06
Merging v4

poovamraj:
In this PR we will be merging the changes we have done for v4 into the master. All the changes for v4 can be found in the Milestone here - https://github.com/auth0/java-jwt/milestone/33


poovamraj:
@stevehobbsdev Everything should be good, as you mentioned all of them are peer reviewed. We should be good to go.

Issue #603, opened 2022-08-15
Make RegisteredClaims values constants


an-kcu:
### Describe the problem you'd like to have solved

I would like to use the values defined in `RegisteredClaims` (and potentially `HeaderParams`) as annotation values, for example to model a JWT (full or just payload) using a custom class. Currently, the values defined in these classes are not marked `final` and therefore cannot be used in annotation values. As far as I can tell, there is no case where they need to be mutable, and it seems to me that these are intrinsically constant values.

### Describe the ideal solution

Make all values defined in `RegisteredClaims` and `HeaderParams` `final` so that they can be used in annotation values. I'm happy to submit a PR, especially if this is the extent of the changes.

### Alternatives and current work-arounds

Either finding an additional library that has these values, or redefining them as constants within my project.

### Additional information, if any

`java-jwt:4.0.0`

poovamraj:
Hi @an-kcu, Thanks for raising this. It is a very valid request and we have raised a PR with this change - https://github.com/auth0/java-jwt/pull/604

I will close this issue for now but feel free to reopen it if you have any more doubts about this.

PR #604, opened 2022-08-22
Make JWT constants final values

poovamraj:
### Changes
We have made all our constants as final values as this would allow us to use them at compile time. This should have been the right way from the beginning as these are constants. This will be a breaking change where we make the fields final so they cannot be mutated. But this behaviour is not expected from the library users, hence we should be able to proceed

### References
https://github.com/auth0/java-jwt/issues/603

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,14 +22,14 @@ commands:
     steps:
       - run: ./gradlew check jacocoTestReport --continue --console=plain
       - codecov/upload
-  run-api-diff:
-    steps:
-      # run apiDiff task
-      - run: ./gradlew apiDiff
-      - store_artifacts:
-          path: lib/build/reports/apiDiff/apiDiff.txt
-      - store_artifacts:
-          path: lib/build/reports/apiDiff/apiDiff.html
+  # run-api-diff:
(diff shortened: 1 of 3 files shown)


poovamraj:
The build is failing because the fields were not final before. This would mean that these values could be changed by the library user. Hence the API Diff job is throwing an error.

But in reality, these values should not be changed and this wouldn't be a breaking change. (Even if users were changing the values of the variables which is highly unlikely) This should be the right way ahead.

jimmyjames:
As the apiDiff build check indicates, this technically is a breaking change. It's also a bug that these fields are not marked final - as the fields correspond to the registered claims defined in [RFC 7519](https://www.rfc-editor.org/rfc/rfc7519#section-4.1), they should not be mutable.

As @poovamraj noted it's not intended that these fields be modified, and such anticipated use is not expected. That said we will need to mark this PR as breaking, clarify the PR description, and note so on the changelog so anyone who does happen to be relying on this bug will be aware of changes required when this is released.

jimmyjames (approved):

jimmyjames (approved):

Issue #597, opened 2022-07-12
Support using Cryptographic providers


aps-dnxt:
Hi y'all

### Describe the problem you'd like to have solved

In my project, I'm required to use an HSM (Hardware Security Module) for computing digital signatures. This particular HSM provides a Java API which includes a java.security.Provider class. This class is tasked with making the HSM specific calls for computing signatures with a given java.security.PrivateKey reference.
The same could be said if we want to use BouncyCastleProvider, another common crypto provider in the Java and .NET crypto communities.
Specifying the signature provider is standard Java, as can be observed in the Signature.getInstance(String algorithm, String provider).
Unfortunately, there's no way in auth0 to pass a provider instance or it's name to the CryptoHelper.

### Describe the ideal solution

Since auth0 is already relying on java.security.Signature class to compute signatures, for me, the ideal solution would be to add a provider parameter to the JWTCreator.Builder#sign(Algorithm algorithm) method to match the java.security.Signature.getInstance(String algorithm, String/Provider provider) method in the Java API.

## Alternatives and current work-arounds

I'm unsure if setting the HSM provider as the provider in position 0 of the providers list would suffice. Other than this, I'm forking auth0 and modifying the code.

### Additional information, if any

I'm using auth0 3.16.0.
I'm also available to implement this and create a pull request with the code. I'm doing it right now..

Regards
Antonio Pedro Santos

aps-dnxt:
Hello again,

I have the code ready but have no permissions to push a new branch ;). I was able to keep the API untouched, as the tests remain unchanged and still compile.

Hope to hear back from you. In the meantime, I'm using my fork from release-4.0.0.

Regards
Antonio Pedro Santo
[provider_support_patch.zip](https://github.com/auth0/java-jwt/files/9097021/provider_support_patch.zip)
s

aps-dnxt:
Yeah, in the previous comment I attached the patch but forgot to say so..

Regards
APS

jimmyjames:
Hi @aps-dnxt, apologies for the delay. Thanks for the request and offer to make a PR. We'll take a look at this and see if this is something that makes sense to add.

aps-dnxt:
Hi @jimmyjames 

In our case it did given that we are using both BouncyCastle's Provider and a Luna HSM (which has it's own provider).
Maybe it's an edge case but it would also provide more flexibility to the lib. Win-win, right?

Regards,
APS

teagy:
[...]

PR #610, opened 2022-09-26
Support specifying the Cryptography Provider by name and instance

aps-dnxt:
### Changes

Treble all sign and verify methods to accept the provider name or a java.security.Provider instance as it provides more flexibility.

* The overloads that take the provider name check if it is loaded into Security and call the overload with the provider instance from Security. If it has not been added, they throw a NoSuchProviderException;
* Old methods without provider, call the new ones with provider as null;
* The methods that take a provider check if null has been passed in. If null, they call the getInstance only with the algorithm as parameter, which will use the lowest ranked provider in Security. This was the default behaviour.

### References

- [support ticket](https://github.com/auth0/java-jwt/issues/597)

### Testing

Tests pass without a hitch. 

- [X] Tests were updated to comply with the changes introduced, namely, the exception expected by Mockito in createSignatureFor and verifySignature. This had to be made because the old methods now call the overload with provider argument and mockito does not follow the implementation. It rather returns immediately.

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,14 +22,14 @@ commands:
     steps:
       - run: ./gradlew check jacocoTestReport --continue --console=plain
       - codecov/upload
-  run-api-diff:
-    steps:
-      # run apiDiff task
-      - run: ./gradlew apiDiff
-      - store_artifacts:
-          path: lib/build/reports/apiDiff/apiDiff.txt
-      - store_artifacts:
-          path: lib/build/reports/apiDiff/apiDiff.html
+  # run-api-diff:
+  #   steps:
+  #     # run apiDiff task
+  #     - run: ./gradlew apiDiff
+  #     - store_artifacts:
+  #         path: lib/build/reports/apiDiff/apiDiff.txt
+  #     - store_artifacts:
+  #         path: lib/build/reports/apiDiff/apiDiff.html
 jobs:
   build:
     docker:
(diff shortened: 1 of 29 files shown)


aps-dnxt:
I'm not able to submit a commit that fixes the check on CircleCI

aps-dnxt (commented):
I believe it's finished!
(famous last words 🥇 )

teagy:
Correct me if I'm wrong but this pull request should allow for the provider to be specified along with the algorithm like so:

# example of specifying BouncyCastle FIPS provider along with algorithm assuming the BC jar is on the classpath
`Mac mac = Mac.getInstance(“HmacSHA256, "BCFIPS");`

aps-dnxt:
> Correct me if I'm wrong but this pull request should allow for the provider to be specified along with the algorithm like so:
> 
> Example of specifying BouncyCastle FIPS provider along with algorithm assuming the BC jar is on the classpath: `Mac mac = Mac.getInstance(“HmacSHA256, "BCFIPS");`

Hi @teagy,

You are correct. But it does not suffice for the BouncyCastle's provider being on the classpath. By giving the name of the provider, Java will still look for it in the installed providers ( ```java.Security.getProvider(String providerName)``` ).
I also introduced the possibility to give an instance of the Provider. In such case, the instance is created by the caller, sufficing the provider's jar to be in the classpath.

Regards
APS

jimmyjames:
👋 @aps-dnxt as I started looking at the changes, I notice that many methods or fields have been moved around. It's hard to decipher exactly what has or hasn't changed in those cases. For the purposes of this PR, would you mind not including any of those types of changes, so we can focus on the actual additions/changes? We can always follow with a separate PR that just reorganizes code if needed.

aps-dnxt:
Hi @jimmyjames 

What I tried to do was to keep the overloads close to the old methods but I get it, it becomes more difficult to read, although easier to write. I can try to move them to the end of the source files if it would make it easier.
In a nutshell, what I did was to empty the old sign and verify methods to use the new that takes a Provider parameter. I also included another one that takes the Provider name, gets the Provider instance from Security and uses the new method.
I'll try to clean it up.
Do you also mean the test cases?

Regards
APS

jimmyjames:
Hi @aps-dnxt, I'm just referring to any changes to the src classes/public APIs. I know it's a bit of ask to reorganize your work, but when I look at the diffs and see fields or methods removed when they're simply moved, it's harder to review. Hope you understand and thanks for your contribution 🙇

jimmyjames:
Closing this PR, but do think the general is a good one, happy to look at a new PR and will put something on our backlog to do this in the future.

aps-dnxt:
Thanks @jimmyjames. I just couldn't manage to find time to tackle this properly. I will in the future open a new one.

PR #628, opened 2022-10-20
Compatible with the original base64 encoded jwt

goofwoof:
So far,The JWT Creator just signed the jwt with urlbase64 encoding and all Algorithms just decode with urlbase64, but if the jwt is signed by someone else with base64. It's hard to verify their jwt.

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWT.java
+++ b/lib/src/main/java/com/auth0/jwt/JWT.java
@@ -28,7 +28,7 @@ public class JWT {
      * @throws JWTDecodeException if any part of the token contained an invalid jwt or JSON format of each of the jwt parts.
      */
     public DecodedJWT decodeJwt(String token) throws JWTDecodeException {
-        return decodeJwt(token, true);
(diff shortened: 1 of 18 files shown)


jimmyjames:
👋 hi @goofwoof! 

Before we look at the specific changes in the PR, can you provide some additional info and steps to reproduce?

jimmyjames:
Taking a quick look at the PR it appears it would add support for non-url-base64 encoding. Per [RFC 7519](https://www.rfc-editor.org/rfc/rfc7519) a JWT must be encoded using Base64 URL encoding (without padding). If there's additional info to share about a specific scenario that this library doesn't support while adhering to RFC-7519, let us know, otherwise this change isn't something this library will support.

jimmyjames:
We will update the javadocs to be more specific. We won't be adding support for non-URL base64 encoding, as doing so would not be compliant with the spec. To workaround legacy systems, you should be able to convert the token to or from url-base64 encoding.

PR #642, opened 2022-11-25
Feature/cleanup

CodeDead:
### Changes

- GIT compliance for some files
- Simpler HashMap declaration in code
- Replaced usage of deprecated `JsonMapper` methods
- Removed unneeded null check when using `instanceof` (because if the value is null, it will return `false` in the `instanceof` check)
- Removed unused imports

### References

None.

### Testing

- [X] This change has been tested on the latest version of Java

### Checklist


jimmyjames (approved):
🎉

jdubjdub:
this `JsonMapper` builder pattern usage (instead of the deprecated `configure` method) breaks anyone using jackson prior to 2.10 :(

CodeDead:
> this `JsonMapper` builder pattern usage (instead of the deprecated `configure` method) breaks anyone using jackson prior to 2.10 :(

Any reason to continue using 4+ year old dependencies? Why not upgrade?

jdubjdub:
enterprise software :D

CodeDead:
> that's what i'm doing, but this seems like a fairly aggressive upgrade, understanding that the `mapper.configure` method was only actually deprecated as of jackson 2.13

Not really sure why this would be considered fairly aggressive or even problematic.

Jackson 2.13 is 2 years old and even the JDK release schedule is on a 6 month basis. Spring updates and deprecates and removes methods with almost every release, which happen more frequently than 2 years.

If anything, waiting 2 years to replace a deprecated method is fairly acceptable in my opinion for a library that is updated on a continual basis.

Issue #622, opened 2022-10-17
Support JWTCreator.Builder to accept Json values


andrewrigas:
I am trying to write a Scala library which basically enhance the functionality on top of [auth0/java-jwt](https://github.com/auth0/java-jwt) library. The problem I need to address is to be able to convert scala types via the encoders in java types, which is not really easy to do because you can have nested types. It will be nice to have an api that would accept a json directly and make some checks that the json passed was in a valid structure for payload and header.

Ideal solution is to be able to expose those 2 functions illustrated below:

```
public Builder withHeader(String headerClaimsJson) 

public Builder withPayload(String payloadClaimsJson)
```

jimmyjames:
:wave: @andrewrigas - I don't think this is something we can reasonably support, as we need to ensure the types/values are of our supported serializable types, and then convert them to a map of string->object. Parsing apart raw JSON strings is something that the client can do and then use existing APIs including the `withPayload(Map<String, ?> payloadClaims)` method.

andrewrigas:
@jimmyjames - Not sure exactly why do you need to convert it back to `Map<String, Object>`? If for example a client passed a valid JSON string structure, then you could convert the rest of the claims to JSON string and merge. The problem with passing a `Map` into the `withPayload(Map<String,?> payloadClaims)` is that I need to provide the map with a key. 

Therefore, the JSON will look like this:
```json 
 {
   "issuer": "foo",
   "subject": "bar",
   "key": {
       "name": "Sam",
       "age": 18
   }
 }
```

Where what I want to achieve is having the client claims in the payload like the example below:
```json
{
   "issuer": "foo",
   "subject": "bar",
   "name": "Sam",
   "age": 18
}
```

jimmyjames:
> Where what I want to achieve is having the client claims in the payload like the example below:

I may be missing something - why can't you just use `name` and `age` as first-level keys in the map passed to `withPayload(Map<String,?> payloadClaims)`?

jimmyjames:
Would something like this work? I think we could add a new method that accepts a JSON string representing the payload, and then delegate to the `withPayload(Map claims)` method to run the existing validation.

```java
public Builder withPayload(String payloadJSON) {
        try {
            Map<String, ?> claims = mapper.readValue(payloadJSON, Map.class);
            return withPayload(claims);
        } catch (JsonProcessingException e) {
            // TODO what exception to throw?
            throw new IllegalArgumentException("Invalid payload JSON", e);
        }
}
```

[...]

PR #643, opened 2022-11-26
Add support for passing json values for header and payload

andrewrigas:
### Changes

Relates to #622

Added support for json values to be pass to the `withHeader` & `withPayload` apis. This should de-couple the library to be only used with java Map and allow easy integration with other languages on JVM.


lib/src/main/java/com/auth0/jwt/JWTCreator.java
```diff
@@ -98,6 +98,24 @@ public Builder withHeader(Map<String, Object> headerClaims) {
...
 
+        /**
+         * Add specific Claims to set as the Header.
+         * If provided json is null then nothing is changed
+         *
+         * @param headerClaimsJson the values to use as Claims in the token's Header.
+         * @return this same Builder instance.
+         * @throws JsonProcessingException if json value has invalid structure
+         */
+        public Builder withHeader(String headerClaimsJson) throws JsonProcessingException {
```
jimmyjames, line 109: Instead of throwing an exception from jackson, let's catch and throw an `IllegalArgumentException` with the nested cause, to be consistent with the other methods and not leak the jackson dependency.

lib/src/main/java/com/auth0/jwt/JWTCreator.java
```diff
@@ -467,6 +485,30 @@ public Builder withPayload(Map<String, ?> payloadClaims) throws IllegalArgumentE
...
+         * If any of the claims are invalid, none will be added.
+         * </p>
+         *
+         * @param payloadClaimsJson the values to use as Claims in the token's payload.
+         * @return this same Builder instance.
+         * @throws IllegalArgumentException if any of the claim keys or null,
+         *                                  or if the values are not of a supported type.
+         * @throws JsonProcessingException if json value has invalid structure
+         */
+        public Builder withPayload(String payloadClaimsJson) throws IllegalArgumentException, JsonProcessingException {
```
jimmyjames, line 502: Same thing here re. the jackson exception.

jimmyjames (changes requested):
Looks good, thanks! Only request is to not throw the jackson exception, and instead wrap it in an `IllegalArgumentException`.

PR #652, opened 2023-01-26
Fix for `exp` claim considered valid if equal to now

### LINKED ISSUE #646: token is still alive, when "exp" = NOW
### Describe the problem
I tried with the code below, but no exception occurred. (TokenExpiredException occurs when `Clock` is +1 second)
This means that token lifetime is `NOW <= "exp"`
```kotlin
// kotlin code
val expiresAt = Instant.now()
val token = JWT.create()
    .withExpiresAt(expiresAt)
    .sign(Algorithm.HMAC256("test"))

val verifier = (JWT.require(Algorithm.HMAC256("test")) as JWTVerifier.BaseVerification)
    .build(Clock.fixed(expiresAt, ZoneId.of("UTC")))

verifier.verify(token)
```

Shouldn't the token lifetime be `NOW < "exp"`?

I think this description is correct.
https://github.com/auth0/java-jwt/blob/master/EXAMPLES.md#datetime-claim-validation

### Environment
- **Version of this library used:** 4.2.1
- **Version of Java used:** 11
- **Other modules/plugins/libraries that might be involved:** Kotlin 1.7.22

jimmyjames:
As discussed in #646, the current date/time must be before the `exp` claim's value.

From the [spec, section 4.1.4](https://www.rfc-editor.org/rfc/rfc7519#section-4.1.4) (emphasis added):

> The processing of the "exp" claim requires that the current date/time MUST be **before** the expiration date/time listed in the "exp" claim.

This change ensures that the current time is before the `exp` claim's value (if present).

Fixes #646

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
+++ b/lib/src/main/java/com/auth0/jwt/JWTVerifier.java
@@ -346,7 +346,7 @@ public final class JWTVerifier implements com.auth0.jwt.interfaces.JWTVerifier {
                     throw new TokenExpiredException(String.format("The Token has expired on %s.", claimVal), claimVal);
                 }
             } else {
-                isValid = assertInstantIsPast(claimVal, leeway, now);
+                isValid = assertInstantIsLessThanOrEqualToNow(claimVal, leeway, now);
                 if (!isValid) {
                     throw new IncorrectClaimException(
                             String.format("The Token can't be used before %s.", claimVal), claimName, claim);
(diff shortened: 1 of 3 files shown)


lib/src/main/java/com/auth0/jwt/JWTVerifier.java
```diff
@@ -356,10 +356,10 @@ private boolean assertValidInstantClaim(String claimName, Claim claim, long leew
         }
 
         private boolean assertInstantIsFuture(Instant claimVal, long leeway, Instant now) {
-            return !(claimVal != null && now.minus(Duration.ofSeconds(leeway)).isAfter(claimVal));
+            return claimVal == null || now.minus(Duration.ofSeconds(leeway)).isBefore(claimVal);
         }
 
-        private boolean assertInstantIsPast(Instant claimVal, long leeway, Instant now) {
+        private boolean assertInstantIsLessThanOrEqualToNow(Instant claimVal, long leeway, Instant now) {
```
jimmyjames, line 362: renamed this method to be clearer

evansims (approved):

PR #774, opened 2026-06-05
Report out-of-range NumericDate claims distinctly from non-numeric ones

### LINKED ISSUE #706: exp claim doesnt support scientific notation
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/java-jwt#readme) and [Examples](https://github.com/auth0/java-jwt/blob/master/EXAMPLES.md), and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://javadoc.io/doc/com.auth0/java-jwt/latest/index.html) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/java-jwt/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

If we have a jwt token, with exp claim value as 
```
"exp": 1.733162101e+26,
```
the library fails to decode the token. We get a JWTDecodeException with error "The claim 'exp' contained a non-numeric date value."

According the RFC on JWT(https://datatracker.ietf.org/doc/html/rfc7519#section-2), a NumericDate should adhere to IEEE Std 1003, according to which scientific notation representation is valid. 

### Reproduction

1. Create a jwt token with exp claim in scientific notation.
2. Try to decode the token. It will fail with "The claim 'exp' contained a non-numeric date value."

### Additional context

_No response_

### java-jwt version

4.4.0

### Java version

17

arpitjain099:
Fixes #706

## What is actually happening

While looking into this I found that in-range scientific notation already decodes correctly. A claim like `{"exp": 1.7e9}` is parsed by Jackson as a `DoubleNode` whose `canConvertToLong()` is true, and `getInstantFromSeconds` already handles it and returns `Instant.ofEpochSecond(1_700_000_000)`.

The value in the report, `1.733162101e+26`, is different: it is about `1.7e26` seconds (a year around `5.5e18`), which overflows `long` and is not representable as a NumericDate or an `Instant`. So it genuinely cannot be decoded to a date. The defect is that `PayloadDeserializer.getInstantFromSeconds` reported it as `The claim 'exp' contained a non-numeric date value`, which is misleading: the value is numeric, it is just out of range.

## Fix

Split the two failure modes in `getInstantFromSeconds`:

- not a number: keep the existing `non-numeric date value` message.
- numeric but overflows `long`: a new, accurate message, `The claim '<name>' value (<value>) is out of the range representable as a NumericDate.`

In-range scientific-notation values continue to decode unchanged. Clamping an absurd overflow value to some wrong year would be worse than a clear rejection, so the out-of-range case is rejected with the precise error.

## Tests

Added to `PayloadDeserializerTest`:

- `shouldGetInstantWhenParsingScientificNotationNode`: `1.7e9` decodes to `Instant.ofEpochSecond(1_700_000_000)` (documents that in-range scientific notation works).
- `shouldThrowOutOfRangeWhenNumericDateExceedsLong`: `1.733162101e+26` raises the new out-of-range message.

The existing non-numeric and normal-integer tests still pass. The directly affected suites (`PayloadDeserializerTest`, `JsonNodeClaimTest`, `PayloadImplTest`, `JWTVerifierTest`, `JWTDecoderTest`) run green. I ran these on JDK 11 because the repo's Gradle wrapper (6.9.2) rejects newer JDKs; CI will exercise the full version matrix.

## Scope

This is narrower than the title might suggest: scientific notation is already supported for representable values, so this PR fixes the misleading error for out-of-range values and adds the missing test coverage rather than adding a new feature. A separate pre-existing edge (a plain integer that passes `canConvertToLong()` but still overflows `Instant.ofEpochSecond`) is orthogonal to this issue and not addressed here.

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
+++ b/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
@@ -74,9 +74,19 @@ class PayloadDeserializer extends StdDeserializer<Payload> {
         if (node == null || node.isNull()) {
             return null;
         }
-        if (!node.canConvertToLong()) {
+        // A NumericDate per RFC 7519 is a JSON number and may be written in scientific notation
+        // (for example 1.7e9). Split the two failure modes so the thrown error is accurate: a
+        // genuinely non-numeric value, versus a numeric value that is out of the range a long can
+        // hold (such as 1.733162101e+26). Previously both produced "non-numeric date value", which
+        // is wrong for the second case and made large/scientific values look unsupported.
+        if (!node.isNumber()) {
             throw new JWTDecodeException(
                     String.format("The claim '%s' contained a non-numeric date value.", claimName));
+        }
+        if (!node.canConvertToLong()) {
+            throw new JWTDecodeException(String.format(
+                    "The claim '%s' value (%s) is out of the range representable as a NumericDate.",
+                    claimName, node.asText()));
         }
         return Instant.ofEpochSecond(node.asLong());
     }
--- a/lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
+++ b/lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
@@ -241,6 +241,34 @@ public class PayloadDeserializerTest {
         assertThat(instant.toEpochMilli(), is(2147493647L * 1000));
     }
 
+    // https://github.com/auth0/java-jwt/issues/706 - a NumericDate may be written in scientific
+    // notation. An in-range value must still resolve to the right instant.
+    @Test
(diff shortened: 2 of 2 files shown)


tanya732:
Thanks for the contribution, @arpitjain099.

This change correctly distinguishes numeric-but-out-of-range values from genuinely non-numeric ones and provides a more accurate error message. The scope also looks appropriate since it is limited to registered-claim deserialization.

LGTM. Thank you!!

tanya732 (approved):

PR #790, opened 2026-07-24
fix: reject NumericDates outside Instant range

### LINKED ISSUE #782: Out-of-range NumericDate claims can throw unchecked DateTimeException during verification
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/java-jwt#readme) and [Examples](https://github.com/auth0/java-jwt/blob/master/EXAMPLES.md), and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://javadoc.io/doc/com.auth0/java-jwt/latest/index.html) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/java-jwt/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

`JWTVerifier.verify(String)` can throw `java.time.DateTimeException` when a registered NumericDate claim (`exp`, `nbf`, or `iat`) is a JSON number that fits in `long` but is outside the range supported by `java.time.Instant`.

The README verification example documents handling invalid tokens through `JWTVerificationException`:

```java
try {
    decodedJWT = verifier.verify(token);
} catch (JWTVerificationException exception) {
    // Invalid signature/claims
}
```

For the out-of-range NumericDate case, the exception escapes as `DateTimeException` before signature verification reaches the usual invalid-token path.

The relevant code path in `4.5.2` / current `master` is:

1. `JWTVerifier.verify(String)` constructs a `JWTDecoder` before verifying the decoded token.
   Reference: https://github.com/auth0/java-jwt/blob/695fd2bea64b8466b872a9d0c2e7019fee7ac86f/lib/src/main/java/com/auth0/jwt/JWTVerifier.java#L450-L452
2. `JWTParser.parsePayload(...)` catches `IOException` from Jackson parsing.
   Reference: https://github.com/auth0/java-jwt/blob/695fd2bea64b8466b872a9d0c2e7019fee7ac86f/lib/src/main/java/com/auth0/jwt/impl/JWTParser.java#L39-L46
3. `PayloadDeserializer.getInstantFromSeconds(...)` checks `canConvertToLong()`, then calls `Instant.ofEpochSecond(node.asLong())`.
   Reference: https://github.com/auth0/java-jwt/blob/695fd2bea64b8466b872a9d0c2e7019fee7ac86f/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java#L72-L81

`9223372036854775807` passes the `long` conversion check but is outside the valid `Instant` epoch-second range. The result is an unchecked `DateTimeException` instead of a `JWTVerificationException` / `JWTDecodeException`.

Expected behavior:

Out-of-range registered NumericDate claims should be rejected through the library's normal invalid-token exception boundary.

Actual behavior:

`JWTVerifier.verify(String)` throws `java.time.DateTimeException`.

### Reproduction

Create the following files in an empty directory.

First create the source directory:

```sh
mkdir -p src/main/java/repro
```

`pom.xml`:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>

  <groupId>repro</groupId>
  <artifactId>java-jwt-numericdate-repro</artifactId>
  <version>1.0.0</version>

  <properties>
    <maven.compiler.source>11</maven.compiler.source>
    <maven.compiler.target>11</maven.compiler.target>
  </properties>

  <dependencies>
    <dependency>
      <groupId>com.auth0</groupId>
      <artifactId>java-jwt</artifactId>
      <version>4.5.2</version>
    </dependency>
  </dependencies>

  <build>
    <plugins>
      <plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>exec-maven-plugin</artifactId>
        <version>3.5.0</version>
      </plugin>
    </plugins>
  </build>
</project>
```

`src/main/java/repro/NumericDateRepro.java`:

```java
package repro;

import com.auth0.jwt.JWT;
import com.auth0.jwt.algorithms.Algorithm;
import com.auth0.jwt.exceptions.JWTVerificationException;

import java.nio.charset.StandardCharsets;
import java.time.DateTimeException;
import java.util.Base64;

public final class NumericDateRepro {
    public static void main(String[] args) {
        String payload = "{\"exp\":9223372036854775807}";
        String token = base64Url("{\"alg\":\"HS256\",\"typ\":\"JWT\"}") + "."
                + base64Url(payload) + ".invalidsig";

        try {
            JWT.require(Algorithm.HMAC256("secret")).build().verify(token);
            System.out.println("unexpected: token accepted");
            System.exit(2);
        } catch (JWTVerificationException expectedBoundary) {
            System.out.println("expected boundary: " + expectedBoundary.getClass().getName());
            System.exit(0);
        } catch (DateTimeException escaped) {
            System.out.println("reproduced unchecked exception: " + escaped.getClass().getName());
            System.out.println("payload=" + payload);
            System.exit(1);
        }
    }

    private static String base64Url(String value) {
        return Base64.getUrlEncoder()
                .withoutPadding()
                .encodeToString(value.getBytes(StandardCharsets.UTF_8));
    }
}
```

Run:

```sh
mvn -q compile exec:java -Dexec.mainClass=repro.NumericDateRepro
```

Observed result:

```text
reproduced unchecked exception: java.time.DateTimeException
payload={"exp":9223372036854775807}
```

Success / failure oracle:

- Current behavior: the command prints `reproduced unchecked exception: java.time.DateTimeException` and exits with status `1`.
- Expected fixed behavior: the command prints `expected boundary: ...JWTVerificationException...` or another library-controlled invalid-token exception and exits with status `0`.

The same behavior can be exercised with `nbf` or `iat` instead of `exp`.

### Additional context

_No response_

### java-jwt version

4.5.2

### Java version

Java 11

ANSHUL-REAL:
Fixes #782

Registered NumericDate claims that fit in a `long` but fall outside `java.time.Instant`'s epoch-second range are now rejected through the library's existing `JWTDecodeException` boundary.

The range check uses `Instant.MIN` and `Instant.MAX` directly and covers both overflow and underflow with regression tests for `Long.MAX_VALUE` and `Long.MIN_VALUE`.

Validation:

- `./gradlew :java-jwt:test --tests com.auth0.jwt.impl.PayloadDeserializerTest -x :java-jwt:compileModuleInfoJava`
- `./gradlew :java-jwt:test -x :java-jwt:compileModuleInfoJava`

### DIFF
--- a/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
+++ b/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
@@ -13,6 +13,7 @@ import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
 
 import java.io.IOException;
+import java.time.DateTimeException;
 import java.time.Instant;
 import java.util.*;
 
@@ -88,7 +89,13 @@ class PayloadDeserializer extends StdDeserializer<Payload> {
                     "The claim '%s' value (%s) is out of the range representable as a NumericDate.",
                     claimName, node.asText()));
         }
-        return Instant.ofEpochSecond(node.asLong());
+        try {
+            return Instant.ofEpochSecond(node.asLong());
+        } catch (DateTimeException e) {
+            throw new JWTDecodeException(String.format(
+                    "The claim '%s' value (%s) is out of the range representable as a NumericDate.",
+                    claimName, node.asText()), e);
+        }
     }
 
     String getString(Map<String, JsonNode> tree, String claimName) {
--- a/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
+++ b/lib/src/main/java/com/auth0/jwt/impl/PayloadDeserializer.java
@@ -13,7 +13,6 @@ import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
 
 import java.io.IOException;
-import java.time.DateTimeException;
 import java.time.Instant;
 import java.util.*;
 
@@ -89,13 +88,13 @@ class PayloadDeserializer extends StdDeserializer<Payload> {
                     "The claim '%s' value (%s) is out of the range representable as a NumericDate.",
                     claimName, node.asText()));
         }
-        try {
-            return Instant.ofEpochSecond(node.asLong());
-        } catch (DateTimeException e) {
+        long seconds = node.asLong();
+        if (seconds < Instant.MIN.getEpochSecond() || seconds > Instant.MAX.getEpochSecond()) {
             throw new JWTDecodeException(String.format(
                     "The claim '%s' value (%s) is out of the range representable as a NumericDate.",
-                    claimName, node.asText()), e);
+                    claimName, node.asText()));
         }
+        return Instant.ofEpochSecond(seconds);
     }
 
     String getString(Map<String, JsonNode> tree, String claimName) {
--- a/lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
+++ b/lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
@@ -270,6 +270,18 @@ public class PayloadDeserializerTest {
     }
 
     @Test
+    public void shouldThrowOutOfRangeWhenNumericDateExceedsInstantRange() {
+        exception.expect(JWTDecodeException.class);
+        exception.expectMessage(
+                "The claim 'key' value (9223372036854775807) is out of the range representable as a NumericDate.");
+
+        Map<String, JsonNode> tree = new HashMap<>();
+        tree.put("key", new LongNode(Long.MAX_VALUE));
+
+        deserializer.getInstantFromSeconds(tree, "key");
+    }
+
+    @Test
     public void shouldGetNullStringWhenParsingNullNode() {
         Map<String, JsonNode> tree = new HashMap<>();
         NullNode node = NullNode.getInstance();
--- a/lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
+++ b/lib/src/test/java/com/auth0/jwt/impl/PayloadDeserializerTest.java
@@ -282,6 +282,18 @@ public class PayloadDeserializerTest {
     }
 
     @Test
+    public void shouldThrowOutOfRangeWhenNumericDatePrecedesInstantRange() {
+        exception.expect(JWTDecodeException.class);
+        exception.expectMessage(
+                "The claim 'key' value (-9223372036854775808) is out of the range representable as a NumericDate.");
+
+        Map<String, JsonNode> tree = new HashMap<>();
+        tree.put("key", new LongNode(Long.MIN_VALUE));
+
+        deserializer.getInstantFromSeconds(tree, "key");
+    }
+
+    @Test
     public void shouldGetNullStringWhenParsingNullNode() {
         Map<String, JsonNode> tree = new HashMap<>();
         NullNode node = NullNode.getInstance();


Zichen1028:
Hi @ANSHUL-REAL,

Thanks for tracking this down — the fix correctly closes the gap from #782, and the repro in the issue was really helpful.

One small gap in test coverage: `Instant`'s range check is symmetric (`seconds < MIN_SECOND || seconds > MAX_SECOND`), but the new test only covers the upper bound (`Long.MAX_VALUE`). It'd be worth adding a mirrored case for the lower bound (e.g. `Long.MIN_VALUE`) to confirm the `catch` handles underflow the same way it handles overflow — right now that's assumed rather than verified.

Separately, and non-blocking for this PR: `getInstantFromSeconds` has now grown three independent guards for essentially the same question ("is this value a valid NumericDate") — `!node.isNumber()` and `!node.canConvertToLong()` from #706, and now this `catch (DateTimeException)` for #782. Each was added reactively when a new edge case surfaced, which makes me think there may be a fourth case waiting somewhere (e.g. if `canConvertToLong()`'s semantics ever shift, or another scientific-notation edge case turns up).

Would it be worth consolidating the range check into a single explicit check against `Instant.MIN`/`Instant.MAX`, rather than relying on `Instant.ofEpochSecond` to throw as a side effect? Something like:

```java
if (!node.canConvertToLong()
        || node.asLong() < Instant.MIN.getEpochSecond()
        || node.asLong() > Instant.MAX.getEpochSecond()) {
    throw new JWTDecodeException(String.format(
            "The claim '%s' value (%s) is out of the range representable as a NumericDate.",
            claimName, node.asText()));
}
return Instant.ofEpochSecond(node.asLong());
```

This keeps the exact same exception types and messages (so no behaviour change for callers), removes the try/catch, and ties the bound directly to `Instant`'s own public constants instead of relying on catching its internal validation. Happy to open this as a separate follow-up issue/PR if you'd rather keep this one focused on #782 — just flagging it here since I ran into it while reviewing.

Zichen1028:
Verified locally in both directions (Long.MAX_VALUE and Long.MIN_VALUE); both are now covered and passing. The Instant.MIN/Instant.MAX approach reads a lot cleaner than the try/catch too. Thanks for making these changes! @ANSHUL-REAL 

> Hi @ANSHUL-REAL,
> 
> Thanks for tracking this down — the fix correctly closes the gap from #782, and the repro in the issue was really helpful.
> 
> One small gap in test coverage: `Instant`'s range check is symmetric (`seconds < MIN_SECOND || seconds > MAX_SECOND`), but the new test only covers the upper bound (`Long.MAX_VALUE`). It'd be worth adding a mirrored case for the lower bound (e.g. `Long.MIN_VALUE`) to confirm the `catch` handles underflow the same way it handles overflow — right now that's assumed rather than verified.
> 
> Separately, and non-blocking for this PR: `getInstantFromSeconds` has now grown three independent guards for essentially the same question ("is this value a valid NumericDate") — `!node.isNumber()` and `!node.canConvertToLong()` from #706, and now this `catch (DateTimeException)` for #782. Each was added reactively when a new edge case surfaced, which makes me think there may be a fourth case waiting somewhere (e.g. if `canConvertToLong()`'s semantics ever shift, or another scientific-notation edge case turns up).
> 
> Would it be worth consolidating the range check into a single explicit check against `Instant.MIN`/`Instant.MAX`, rather than relying on `Instant.ofEpochSecond` to throw as a side effect? Something like:
> 
> ```java
> if (!node.canConvertToLong()
>         || node.asLong() < Instant.MIN.getEpochSecond()
>         || node.asLong() > Instant.MAX.getEpochSecond()) {
>     throw new JWTDecodeException(String.format(
>             "The claim '%s' value (%s) is out of the range representable as a NumericDate.",
>             claimName, node.asText()));
> }
> return Instant.ofEpochSecond(node.asLong());
> ```
> 
> This keeps the exact same exception types and messages (so no behaviour change for callers), removes the try/catch, and ties the bound directly to `Instant`'s own public constants instead of relying on catching its internal validation. Happy to open this as a separate follow-up issue/PR if you'd rather keep this one focused on #782 — just flagging it here since I ran into it while reviewing.

